PR #47, opened 2024-12-17
Test readthedocs

BradleySappington:
update some read the docs formatting

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -2,6 +2,7 @@ Documentation for STPSF
 ===============================
 
 **(Formerly WebbPSF, for versions < 2.0)**
+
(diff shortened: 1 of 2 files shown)


marcio-melendez (commented):
When I do a preview of index, I see most of the document as expected but "contents" looks bad. However, I am not sure if this is because I am using "preview". As long as you don't see it like that then it is fine. 
Screenshot 2024-12-17 at 1 44 08 PM

york-stsci (approved):
Looks good. Adding in note that readthedocs doesn't require blank lines before sub-lists, but I don't think that's an issue because I don't think there are any plans to port the changes anywhere.

BradleySappington:
@obi-wan76 - all looks normal on Readthedocs :)

PR #50, opened 2024-12-18
tweak/enhance auto data install

mperrin:
Some enhancements building on the very nice data installation function added in https://github.com/spacetelescope/webbpsf/pull/932 by @WilliamJamieson :

- If the data was already installed into the default data path by some prior call to `auto_download_stpsf_data`, then there's no need to emit any warning, just use that automatically installed data. 
- Add a back compatibility function `get_webbpsf_data_dir` which is a deprecated equivalent for `get_stpsf_data_dir`. This is simply to allow any existing webbpsf user code that calls `webbpsf.utils.get_webbpsf_data_dir()` to continue to work without error. The deprecation is marked using the recently-added standard library `@deprecated` decorator, or the back ported equivalent depending on Python version. 
- Update the installation docs to include the information about the automatic data install, and no longer say that it's required to manually install the data.


mperrin:
hmm, a side issue in this: I was hoping to use the recently-added standard library `@deprecated` decorator, but that's not yet in all the versions of Python we support. And the back port of that decorator in library `typing_extensions` isn't installed by default either. We could add that to the requirements but that feels like overkill.   

Mostly I would like to avoid unnecessary effort writing our own deprecation warning machinery -- this very much feels like something we should be able to get from upstream somewhere. 

Let me think about an alternative way to do that...

mperrin:
OK, I think it will work to, in `webbpsf.__init__.py`, after doing the sys.modules magic, then we can do

```
webbpsf.utils.get_webbpsf_data_path = stpsf.utils.get_stpsf_data_path
```
and I think that works, based on some testing here. And, no particular deprecation warning is needed in this case, because it'll be covered under the overall deprecation warning at the top of webbpsf.

BradleySappington:
> OK, I think it will work to, in `webbpsf.__init__.py`, after doing the sys.modules magic, then we can do
> 
> ```
> webbpsf.utils.get_webbpsf_data_path = stpsf.utils.get_stpsf_data_path
> ```
> 
> and I think that works, based on some testing here. And, no particular deprecation warning is needed in this case, because it'll be covered under the overall deprecation warning at the top of webbpsf.

 get_webbpsf_data_path will not actually be used once webbpsf2.0.0 is released.  Webbpsf will directly call get_stpsf_data_path.

docs/installation.rst
```diff
@@ -61,21 +53,22 @@ Stsynphot is an optional dependency, but is highly recommended. Its installation
...
 
-.. (If you install via Conda, the data files are automatically installed, in
-    which case you can skip this section.) [uncomment once conda installation is
-    available again]
+Files containing such information as the JWST pupil shape, instrument throughputs, and aperture positions are distributed separately from STPSF. These sum to about 100 MB in size.
 
-Files containing such information as the JWST pupil shape, instrument throughputs, and aperture positions are distributed separately from STPSF. To run STPSF, you must download these files and tell STPSF where to find them using the ``STPSF_PATH`` environment variable.
+STPSF will now **automatically download and install these files** the first time it is run.  By default these will be downloaded and installed into a folder ``$HOME/data/stpsf-data``  within the user's home directory. This directory will be created automatically if it doesn't exist already. If this path is acceptable to you, then no further action is needed on your part, and it'll all work automatically.
```
BradleySappington, line 61: Good addition!  Thanks for cleaning up the docs!

stpsf/utils.py
```diff
@@ -216,30 +220,32 @@ def auto_download_stpsf_data():
...
         if path is None:
-            message = (
-                'Environment variable $STPSF_PATH is not set!\n'
-                f'{MISSING_STPSF_DATA_MESSAGE} searching default location..s'
-            )
-            warnings.warn(message)
-            path = auto_download_stpsf_data()
+
+            # Check if the data were already downloaded to the default path
+            default_path = get_default_data_path()
```
BradleySappington, line 239: This change isn't necessary as this logic already exists within `auto_download_stpsf_data `(which also sets the STPSF_PATH which is necessary)

`auto_download_stpsf_data()` will make a new default directory (only if one doesn't exist), and download the data (only if the directory is empty).

BradleySappington (changes requested):
installation.rst changes are great
Please undo the utils.py changes for reasons in the comment

mperrin:
> get_webbpsf_data_path will not actually be used once webbpsf2.0.0 is released. Webbpsf will directly call get_stpsf_data_path.

My point is about _existing user code that calls that function directly_. That function is currently part of the public API of webbpsf, and I know for a fact there are existing notebooks and downstream scripts that call it directly. Not a lot but nonzero. I completely agree that all the internal code you updated to call get_stpsf_data_path. But again I’m trying to make sure we don’t break some existing notebooks or script that calls get_webbpsf_data_path directly.

stpsf/utils.py
mperrin: My point here was intentionally refactoring to allow _checking for a prior successful call_ to auto_download_stpsf_data, prior to displaying the big giant warning message about trying to download the data. If the data has already been downloaded successfully, it’s incorrect to display a warning that says “stpsf will try to download the data”. 

The current code will display the warning text every time the package is imported in a new Python session, even if the data is already indeed present and fine in the default directory. That’s excess warnings for no benefit to the user.

BradleySappington:
> > get_webbpsf_data_path will not actually be used once webbpsf2.0.0 is released. Webbpsf will directly call get_stpsf_data_path.
> 
> My point is about _existing user code that calls that function directly_. That function is currently part of the public API of webbpsf, and I know for a fact there are existing notebooks and downstream scripts that call it directly. Not a lot but nonzero. I completely agree that all the internal code you updated to call get_stpsf_data_path. But again I’m trying to make sure we don’t break some existing notebooks or script that calls get_webbpsf_data_path directly.

AH! Yes thats a good call.  Overlooked that we should verify there are no other public API calls that specifically have "webbpsf" as part of the naming convention.

(just confirmed that its only auto_download_webbpsf_data and get_webbpsf_data_path)

stpsf/utils.py
BradleySappington: I see your concern (deleted my previous reply because I misunderstood what download message you were referring to)

I think its important that the `Environment variable $STPSF_PATH is not set!` warning is still displayed.  

May just make more sense to move the `MISSING_STPSF_DATA_MESSAGE` inside `auto_download_stpsf_data` inside the `if not any(default_path.iterdir()):` codeblock.

marcio-melendez:
@BradleySappington  Did we ever come back to this PR? is there an update? 
Thanks!

stpsf/utils.py
BradleySappington: @mperrin and @obi-wan76 - this is the only lingering discussion on this PR.  I Trust your opinion if you think the `Environment variable $STPSF_PATH is not set!` should or should not remain in the code.

mperrin: What I'm wondering is, for the end user, what benefit (if any) is there to having this message pop up, if the user doesn't have to set that any more because now there's this automatic install? 

In other words, doesn't it make it easier for the users, and more automatic/seamless/easy in general, to have the _default recommended behavior_ become "the data gets installed automatically and the user doesn't have to think about it and the code just works"? There's still the option to manually install the data someplace and set the environment variable if it is so needed, but probably it would be only a tiny fraction of users who would need to do that.  

@obi-wan76 what do you think is better for a user-support perspective?

marcio-melendez: I am always in favor of adding as many explicit messages to the user as possible. I see Marshall's point that the whole reason to have an automate process for the setup is for the user to be transparent to what's happening, therefore no need for extra information. However, I think that in this case the code is about to download data into a local folder, thus, I think it is appropriate to inform the user. In fact, we could add to the message "we are attempting to download the data".

mperrin: To be clear, I'm definitely in favor of having a message that informs the user that it is attempting to download the data! But not every single time. 


Maybe I have not been clear enough in general about what I'm thinking, so let me try to explain. 

Here is what happens now, if you try to run this without any data files installed or environment variable set: 
```

In [1]: import stpsf

In [2]: nrc = stpsf.NIRCam()
/Users/mperrin/Dropbox (Personal)/Documents/software/stpsf/stpsf/utils.py:243: UserWarning: Environment variable $STPSF_PATH is not set!

 ***********  ERROR  ******  ERROR  ******  ERROR  ******  ERROR  ***********
 *                                                                          *
 *  STPSF requires several data files to operate.                         *
 *  These files could not be located automatically at this time, or this    *
 *  version of the software requires a newer set of reference files than    *
 *  you have installed.  For more details see:                              *
 *                                                                          *
 *        https://stpsf.readthedocs.io/en/stable/installation.html        *
 *                                                                          *
 *  under "Installing the Required Data Files".                             *
 *  STPSF will not be able to function properly until the appropriate     *
 *  reference files have been downloaded to your machine and installed.     *
 *                                                                          *
 ****************************************************************************
 searching default location..s
  warnings.warn(message)
/Users/mperrin/Dropbox (Personal)/Documents/software/stpsf/stpsf/utils.py:198: UserWarning: STPSF data files not found in default location, attempting to download them now...
  warnings.warn("STPSF data files not found in default location, attempting to download them now...")

In [3]: psf = nrc.calc_psf()
# that works to compute a PSF, and there is no problem
```

So, that's OK but not great because it says "attempting to download them now" but then doesn't also say "the download was successful" after that. It should say the download is successful if that is the case. Also, having the error message be a big ERROR ERROR that shouts so intensely is overdoing it, because actually it's not a big deal, since the auto download just takes care of it. No user action required. 

Now let's look at what happens the next time the user starts a new Python session later and imports stpsf again the next time. 

```

In [1]: import stpsf

In [2]: nrc = stpsf.NIRCam()
/Users/mperrin/Dropbox (Personal)/Documents/software/stpsf/stpsf/utils.py:243: UserWarning: Environment variable $STPSF_PATH is not set!

 ***********  ERROR  ******  ERROR  ******  ERROR  ******  ERROR  ***********
 *                                                                          *
 *  STPSF requires several data files to operate.                         *
 *  These files could not be located automatically at this time, or this    *
 *  version of the software requires a newer set of reference files than    *
 *  you have installed.  For more details see:                              *
 *                                                                          *
 *        https://stpsf.readthedocs.io/en/stable/installation.html        *
 *                                                                          *
 *  under "Installing the Required Data Files".                             *
 *  STPSF will not be able to function properly until the appropriate     *
 *  reference files have been downloaded to your machine and installed.     *
 *                                                                          *
 ****************************************************************************
 searching default location..s
  warnings.warn(message)
```

It still prints the exact same error message. But actually the reference data are _already downloaded_.  There's no user action required, they can just use the code to make PSF and everything is fine. So there is no benefit to telling the user "ERROR  ******  ERROR  ******  ERROR " in that case! It would just cause confusion.

mperrin: Now let's look at what happens with this PR. 

First time a user runs this without the data, it's basically the same, but it does also say that the data was installed:
```

In [1]: import stpsf

In [2]: nrc = stpsf.NIRCam()
/Users/mperrin/Dropbox (Personal)/Documents/software/stpsf/stpsf/utils.py:249: UserWarning: Environment variable $STPSF_PATH is not set!

 ***********  ERROR  ******  ERROR  ******  ERROR  ******  ERROR  ***********
 *                                                                          *
 *  STPSF requires several data files to operate.                         *
 *  These files could not be located automatically at this time, or this    *
 *  version of the software requires a newer set of reference files than    *
 *  you have installed.  For more details see:                              *
 *                                                                          *
 *        https://stpsf.readthedocs.io/en/stable/installation.html        *
 *                                                                          *
 *  under "Installing the Required Data Files".                             *
 *  STPSF will not be able to function properly until the appropriate     *
 *  reference files have been downloaded to your machine and installed.     *
 *                                                                          *
 ****************************************************************************
 searching default location.
  warnings.warn(message)
/Users/mperrin/Dropbox (Personal)/Documents/software/stpsf/stpsf/utils.py:201: UserWarning: STPSF data files not found in default location, attempting to download them now...
  warnings.warn("STPSF data files not found in default location, attempting to download them now...")
/Users/mperrin/Dropbox (Personal)/Documents/software/stpsf/stpsf/utils.py:214: UserWarning: STPSF data files downloaded and installed to /Users/mperrin/data/stpsf-data.
  warnings.warn(f"STPSF data files downloaded and installed to {default_path}.")

In [3]: psf = nrc.calc_psf()

In [4]:
```

Then they finish what they're doing, and some time later make a new Python session. This time, since the code can tell that the data was already downloaded successfully by the auto-download feature, there's no need to print any errors or warnings at all. It just works. 

```

In [1]: import stpsf

In [2]: nrc = stpsf.NIRCam()

In [3]: psf = nrc.calc_psf()

```
That seems better to me, for 2nd and subsequent uses after the data was already downloaded. There is no benefit to printing a false-positive error message that says "STPSF will not be able to function until the reference files are downloaded" when they have already been downloaded and the code will function just fine.

BradleySappington: @mperrin - I believe there is disconnect on my feedback as well.  I agree with you that not having the massive error sign repeat itself is a good thing.  There is a very small error directly above it. `Environment variable $STPSF_PATH is not set!`

That is the error that I'm speaking to.  It was moved along with the massive error in this PR.  I suggest that it should persist because it is part of our installation process, and the user should know that they have overlooked setting it correctly (despite our very clever download mechanisms taking care of stuff for them).  

So all my comments above are saying is that the following line should stick around as it indicates setup was not followed:
`/Users/mperrin/Dropbox (Personal)/Documents/software/stpsf/stpsf/utils.py:249: UserWarning: Environment variable $STPSF_PATH is not set!`


And this code block below shouldn't (and in all honesty should probably be calmed down a bit)

```
 ***********  ERROR  ******  ERROR  ******  ERROR  ******  ERROR  ***********
 *                                                                          *
 *  STPSF requires several data files to operate.                         *
 *  These files could not be located automatically at this time, or this    *
 *  version of the software requires a newer set of reference files than    *
 *  you have installed.  For more details see:                              *
 *                                                                          *
 *        https://stpsf.readthedocs.io/en/stable/installation.html        *
 *                                                                          *
 *  under "Installing the Required Data Files".                             *
 *  STPSF will not be able to function properly until the appropriate     *
 *  reference files have been downloaded to your machine and installed.     *
 *                                                                          *
 ****************************************************************************
```

mperrin: Aha, got it. I think we have succeeded in communication ;-) 

Let me take a shot at rewriting and calming down that error message too, I'll push a change for that later. That level of error made sense when it truly did require the user to do something manually or else nothing would work, but that's not the case any more.

BradleySappington: Awesome, I'll resolve this convo and await a tag saying you're ready for a final review or merge :)

BradleySappington (approved):

PR #51, opened 2024-12-18
For coronagraph masks, add reference information on alignment per Lyot mask, and also MIRI FQPM central wavelengths

mperrin:
**[relocated from https://github.com/spacetelescope/webbpsf/pull/675]**

This PR implements a capability to have default mask alignment options (`pupil_shift_x, pupil_shift_y, pupil_rotation`) defined per each coronagraph mask. This is intended to allow WebbPSF to include measured in-flight information on the various masks. 

Implementation details: 
- in `constants.py` add a dict which gives values of these options per each mask, or else None to leave undefined
- in `webbpsf_core.py`, update code to look up and use those values. 

_WORK IN PROGRESS, NOT YET COMPLETE, NOT READY TO MERGE_

### DIFF
--- a/stpsf/constants.py
+++ b/stpsf/constants.py
@@ -416,10 +416,16 @@ INSTRUMENT_IFU_BROADENING_PARAMETERS = {
 
 # Alignment information about instrument internal pupil masks (
 INSTRUMENT_PUPIL_MASK_DEFAULT_POSITIONS = {
-    'NIRCam_MASKSWB': {'pupil_shift_x': None, 'pupil_shift_y': None, 'pupil_rotation': None},
-    'NIRCam_MASKLWB': {'pupil_shift_x': None, 'pupil_shift_y': None, 'pupil_rotation': None},
-    'NIRCam_MASKRND_SW': {'pupil_shift_x': None, 'pupil_shift_y': None, 'pupil_rotation': None},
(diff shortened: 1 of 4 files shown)


mperrin:
Make sure to also see the comment thread and discussion at https://github.com/https://github.com/spacetelescope/webbpsf/pull/675

mperrin:
Finally getting back to this after a long while, during a collaborative visit with MIRI collaborators from Paris Observatory.

mperrin:
FYI @kdlawson

Also FYI @juliengirard @JarronL @aggle 

Kellen I don't think I can formally tag you as reviewers on this code since you're not in the spacetelescope GitHub organization, but I will be grateful if you can give this a careful look over, please

mperrin:
Now tested and debugged.  Several additional changes are now in this PR: 

1. Also writes new FITS headers keywords for PUPLSHFX, PUPLSHFY, PUPL_ROT to record the values used in a given calculation
2. Relevant unit tests for MIRI and NIRCam coronagraphy calcs are updated to check those keywords to verify that the test calculation correctly got the default value from the constants file. 
3. For MIRI, this PR also implements lookup constants for the mask central wavelengths, which may not exactly match the notional design wavelength. This is in particular definitely true for MASK1140.  The value of this is saved as a new header keyword FQPMWAVE

kdlawson:
@mperrin Looks good to me! The defaults all match what I have. Adding the values used to the headers is also a really helpful book-keeping feature. Thanks for putting this together!

PR #57, opened 2025-01-10
example notebook on how to transition from webbpsf to stpsf

marcio-melendez:
This is an example notebook about a typical usage when transitioning from webbpsf to stpsf. Comments are welcome!


marcio-melendez:
Hi @BradleySappington and @york-stsci,
This notebook is inteneded to facilitate the transition from webbpsf to stpsf for current webbpsf user. Please, let me know if this make sense to you and if you would like to add more technical details. I did try to keep it as simple as possible. 
Thanks!

BradleySappington (commented):
This looks good Marcio.  Thanks for taking the time to put this together.  I think the only addition i would make is a quick installation update:

It should probably include that if the user is installing WebbPSF in their library via requirements.txt or pyproject.toml, they should completely replace WebbPSF with STPSF (we don't want users to accidentally keep WebbPSF in their install)

marcio-melendez:
> This looks good Marcio. Thanks for taking the time to put this together. I think the only addition i would make is a quick installation update:
> 
> It should probably include that if the user is installing WebbPSF in their library via requirements.txt or pyproject.toml, they should completely replace WebbPSF with STPSF (we don't want users to accidentally keep WebbPSF in their install)

I added a line in the first cell, let me know if that works. I wanted to make a statement about  webbpsf and stpsf version numbers as we progress in time, so people doesn't get confuse that if installing webbpsf they may see the same version number of webbpsf despite the fact that they have the latest version of the code via stpsf. Do you have any suggestions that doesn't confuse people :)

BradleySappington:
A few updates consolidated into the initial paragraph

```
To reflect its broader support for Roman as well as James Webb, WebbPSF functionality has now migrated to STPSF (Space Telescope PSF). STPSF functionality is currently the same as WebbPSF, only the name is different. 

For users who have not switched to using STPSF:
WebbPSF is now an alias of STPSF and is running code from the STPSF library. Therefore, the versioning of WebbPSF will not be parallel to the released versions of STPSF.  WebbPSF will automatically include the newest STPSF changes.

For users ready to switch to STPSF:
If you are ready to make the change to STPSF, this notebook will help you explore the best options in order to transition your WebbPSF scripts and codes into STPSF. Generally speaking there are two options:

import stpsf as webbpsf. This can be done at the beginning of your code and it will keep your code as-is, preserving WebbPSF syntax to function calls.  This is beneficial if you just want to do a quick change to your codebase.  However, a full rename (as outlined below), will lead to a more legible and healthy codebase.

import stpsf. This is the preferred method.  In this case you will fully transition to the STPSF usage, thus, you need to replace all the WebbPSF calls to STPSF.
Note that in both cases you'll be required to install STPSF, including downloading the STPSF data files and setting a STPSF_PATH environment variable. See instructions under [Requirements & Installation](https://stpsf.readthedocs.io/en/latest/installation.html#requirements-installation). 
WebbPSF is no longer a requirement and can be removed from your environment.  Any references to WebbPSF in your library via requirements.txt or pyproject.toml, should be replaced with STPSF.
```

marcio-melendez:
> A few updates consolidated into the initial paragraph
> 
> ```
> To reflect its broader support for Roman as well as James Webb, WebbPSF functionality has now migrated to STPSF (Space Telescope PSF). STPSF functionality is currently the same as WebbPSF, only the name is different. 
> 
> For users who have not switched to using STPSF:
> WebbPSF is now an alias of STPSF and is running code from the STPSF library. Therefore, the versioning of WebbPSF will not be parallel to the released versions of STPSF.  WebbPSF will automatically include the newest STPSF changes.
> 
> For users ready to switch to STPSF:
> If you are ready to make the change to STPSF, this notebook will help you explore the best options in order to transition your WebbPSF scripts and codes into STPSF. Generally speaking there are two options:
> 
> import stpsf as webbpsf. This can be done at the beginning of your code and it will keep your code as-is, preserving WebbPSF syntax to function calls.  This is beneficial if you just want to do a quick change to your codebase.  However, a full rename (as outlined below), will lead to a more legible and healthy codebase.
> 
> import stpsf. This is the preferred method.  In this case you will fully transition to the STPSF usage, thus, you need to replace all the WebbPSF calls to STPSF.
> Note that in both cases you'll be required to install STPSF, including downloading the STPSF data files and setting a STPSF_PATH environment variable. See instructions under [Requirements & Installation](https://stpsf.readthedocs.io/en/latest/installation.html#requirements-installation). 
> WebbPSF is no longer a requirement and can be removed from your environment.  Any references to WebbPSF in your library via requirements.txt or pyproject.toml, should be replaced with STPSF.
> ```

@BradleySappington thanks, I just added your updates..

BradleySappington:
@obi-wan76 I think the only other thing I would do is put `import stpsf` section first, as thats the preferred method, aside from that I'm good with this. Nice work!

marcio-melendez:
> @obi-wan76 I think the only other thing I would do is put `import stpsf` section first, as thats the preferred method, aside from that I'm good with this. Nice work!

Done. Thanks!

BradleySappington (approved):
Nice work Marcio!

PR #59, opened 2025-01-29
small change to remove hard coded field point in trending

marcio-melendez:
small change to remove hard coded field point in the TA image comparison function


BradleySappington (approved):
typically I favor f strings for manipulation like this, however due to the length of the variables I actually consider this more legible.  LGTM

PR #66, opened 2025-02-11
Update readthedocs for webbpsf_to_stpsf notebook

BradleySappington:
Update the webbpsf_to_stpsf to not contain the header error
Include webbpsf_to_stpsf as subheader link at top of index page
Include Webbpsf_to_stpsf on left side index
See yellow circles:
Screenshot 2025-02-11 at 9 56 29 AM

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1,7 +1,6 @@
 Documentation for STPSF
 ===============================
-
-**(Formerly WebbPSF, for versions < 2.0)**
+`(Formerly WebbPSF, for versions < 2.0) <webbpsf_to_stpsf.ipynb>`_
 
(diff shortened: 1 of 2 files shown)


mperrin:
Just looking at your screenshot, I see an "import stpsf" in the sidebar, the line right under the circled one. That also looks like some kind of header error, since that doesn't make sense to have as a subheading?

BradleySappington:
> Just looking at your screenshot, I see an "import stpsf" in the sidebar, the line right under the circled one. That also looks like some kind of header error, since that doesn't make sense to have as a subheading?

Good catch! thats a link to another header in the notebook.  I'll take care of it.

marcio-melendez (approved):
Looks good, thanks!

BradleySappington:
Failing CI tests now fixed in development

PR #67, opened 2025-02-13
Fix NIRISS bug by adding conditional statement to handle PUPIL filters 

tonysohn:
Added a conditional statement in `match_data.py` for NIRISS.

NIRISS filters are actually in both the filter and pupil wheels. When FILTER=CLEAR, the filter is in the PUPIL wheel. Code has been tested on a few NIRISS images. Before the fix, `setup_sim_to_match_file` was incorrectly assuming Pupil plane mask = GR700XD for observations using FILTER=CLEAR / PUPIL=F090W (or any other filter). After the fix, I confirmed that the code identifies the correct filter.


marcio-melendez (approved):
This looks good, thanks!

PR #68, opened 2025-02-17
minor cleanup and consolidation

zacharyburnett:
while looking into ways to embed data into the built wheel, I found a few minor things that could be  cleaned up:
- rename `readthedocs.yml` -> `.readthedocs.yaml` (RTD has indicated that they will eventually deprecate the old name)
- use latest author list in `CITATION.cff` (I preserved the order from `README.md` but wasn't sure if that was the desired order)
- move data globs from `MANIFEST.in` -> `pyproject.toml` `tools.setuptools.package-data`


zacharyburnett:
I see that `requirements.txt` provides upper pins for functional dependencies, creating a stable environment; as it is, however, it appears that the dependencies installed with a `pip install` using `pyproject.toml` will always be equal or later versions than those made from `requirements.txt`, since `[project.dependencies]` does not have upper pins. 

However, in `tox.ini`, the env created from `pyproject.toml` has the `-stable` label, and the one created from `requirements.txt` is labeled `-latest`. Shouldn't these two names be flipped?

BradleySappington:
@zacharyburnett - at this point the requirements.txt is associated with dependabot updates, the upper limit is there solely to have dependabot let us know when the 'Y' version of x.Y.z is updated for us to verify compatibility.  That said, having no upper limit on the pyproject.toml essentially does the same thing (although the pyproject highest available versions should essentially always be in line with the requirements.txt upper limit due to dependabot).  I'm confident there's a solution to handle those checks but that hasn't been a priority yet for this project.  If you have suggestions please let us know.

marcio-melendez:
@BradleySappington do you agree with this PR? can this be merger? Thanks!

BradleySappington (approved):
LGTM

PR #70, opened 2025-02-18
update release.rst for handling Webbpsf support on box

BradleySappington:
in preparing for a webbpsf release, it became apparent that we need a specific step to account for webbpsf data on box.  adding instructions for future stpsf releases while webbpsf is still being supported.

### DIFF
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -35,6 +35,12 @@ Releasing new data packages
     #. This will change the name of ``minimal-stpsf-data-LATEST.tar.gz`` to be what you just uploaded, rename the file back to ``minimal-stpsf-data-LATEST.tar.gz``
     #. Upload to Box a separate version of ``minimal-stpsf-data-#.#.#.tar.gz`` shared data folder for future storage.
     #. Verify the shared link of ``stpsf-data-latest.tar.gz`` is the same that exists in ``docs/installation.rst`` ("copy shared link" then "link settings")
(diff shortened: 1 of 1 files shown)


mperrin:
Just thinking about options for a second. I don't quite understand the logic of this; can you explain under what circumstances this is needed? For future versions of "webbpsf" which will be using the sys.modules redirect to invoke `stpsf`, what is the use case for having a separate `minimal-webbpsf-data-#.#.#.tar.gz` file for future versions? 

The "minimal" tar file was originally motivated to have a smaller download for the CI, but (a) we switched to using the file cache on Github for that instead of re-downloading every time, and (b) we decided not to keep running CI on Git anyway for this. 

So either I am missing something (entirely possible), or maybe we can convince ourselves it's not necessary to do this. 


To be clear, we want to not break existing uses of webbpsf during this transitional period -- but I don't think releasing new versions of the data file tar.gz is required to do that.

BradleySappington:
There are current CI workflows in place that currently require the directory `/webbpsf-data/`
Our latest downloads from box have everything contained in a directory `/stpsf-data/`

This is to act as a protection for those who have not yet done any conversion at all.  If we release a new version of STPSF with a higher min version, and webbpsf then uses that higher min version, unchanged CI infrastructure will need to grab that data and have it labelled the legacy way.

This came up as @obi-wan76 and I were testing fresh installs, and I updated the webbpsf-latest.tar.gz shared info to be the same as the stpsf latest.  This ended up breaking an internal pipeline for a different team due to the directory name change within the tar file.

mperrin:
Ah I see, it's somebody else's CI workflow that's the issue. Got it, thanks. 👍

marcio-melendez (approved):
just added information, looms good

PR #71, opened 2025-02-19
Change Roman detector naming to support WFIxy 

### LINKED ISSUE #3: Change Roman detector naming from SCAxy to WFIxy
For consistency with other Roman simulation tools as well as with the Roman pipeline, we would like to specify which detector to use with the ```WFIxy``` string instead of ```SCAxy```.

For example, users would use the following going forward:
```wfi.detector = "WFI01"```

### PR DESCRIPTION
**Pull request by Skyhawk172**
This PR addresses https://github.com/spacetelescope/stpsf/issues/3

Given that much of the underlying data files (pupils, zernikes) and implementation relies on detector naming containing SCA, the proposed implementation allows for users to specify any detector with the "WFIxy" string using the detector setter, which then gets converted back to SCA for internal and backward compatibility. 

Note that the header information as well as some verbose statements are also being updated to specify WFIxy.

### DIFF
--- a/stpsf/roman.py
+++ b/stpsf/roman.py
@@ -374,7 +374,11 @@ class RomanInstrument(stpsf_core.SpaceTelescopeInstrument):
         super()._get_fits_header(result, options)
         result[0].header['DETXPIXL'] = (self.detector_position[0], 'X pixel position (for field dependent aberrations)')
         result[0].header['DETYPIXL'] = (self.detector_position[1], 'Y pixel position (for field dependent aberrations)')
-        result[0].header['DETECTOR'] = (self.detector, 'Detector selected')
+
+        if self.telescope == 'Roman':
+            result[0].header['DETECTOR'] = (self.detector.replace("SCA", "WFI"), 'Detector selected')
+        else:
+            result[0].header['DETECTOR'] = (self.detector, 'Detector selected')
 
     def _calc_psf_format_output(self, result, options):
         """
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: stpsf/roman.py [hunk 1]
```diff
@@ -374,7 +374,11 @@ def _get_fits_header(self, result, options):
         super()._get_fits_header(result, options)
         result[0].header['DETXPIXL'] = (self.detector_position[0], 'X pixel position (for field dependent aberrations)')
         result[0].header['DETYPIXL'] = (self.detector_position[1], 'Y pixel position (for field dependent aberrations)')
-        result[0].header['DETECTOR'] = (self.detector, 'Detector selected')
+
+        if self.telescope == 'Roman':
```
mperrin, line 378: Minor: I don't think this if statement is needed, since this is the RomanInstrument class. By construction it should only ever be called for cases with Roman, not JWST.

mperrin (commented):
This looks good, but I have an overall question about whether in the long run it would simplify things if we simply updated the code internals to use "WFIxy" instead of leaving them with "SCAxy". If the Roman-mission-preferred nomenclature uses "WFIxy", then why not switch to that here too? It might be a few more lines of code to change now but might simplify in the long run?   Just something to think about; either way is probably reasonable. 

In any case I would recommend also adding a test function into `test_roman.py` that would test that it works correctly to set the detector using either naming convention, and the value is set as expected.

stpsf/roman.py
Skyhawk172: I agree - this is not necessary given that we're within the RomanInstrument class.

Skyhawk172:
I contemplated switching everything to WFIxy, but then realized that our pupil and Zernike files all rely on the SCAxy string for picking the right file and/or aberrations. 

We could change these files to use WFIxy going forward, or we could leave the files unchanged and add the hooks to pick the right file based on SCAxy. I'll take another look.

Skyhawk172:
Addresses in PR #100

PR #74, opened 2025-02-27
Rename two notebooks for webbpsf->stpsf

mperrin:
Looks like these got missed in the name migration fun. 

It's worth renaming these ASAP since one of these is linked to from the docs front page, and the link text got updated to use "STPSF" already even though the file is not yet there. So this link on the docs homepage currently 404 errors: 

Screenshot 2025-02-27 at 9 21 20 AM


BradleySappington (approved):
Good find!

PR #75, opened 2025-02-27
update notebook path

BradleySappington:
following up on #74 
Using "develop" instead of "stable" in path name so changes will take effect on readthedocs immediately.


mperrin (approved):
Another good catch!

york-stsci (approved):
Looks good

PR #80, opened 2025-03-18
removing the filter keyword in tar extractall

### LINKED ISSUE #79: tarfile extract keyword only valid for python 3.12+
The keyword ``filter="fully_trusted"`` appears to have been introduced in python 3.12, so currently throws an error in previous versions.
https://github.com/spacetelescope/stpsf/blob/9a0094d5a855cf3d74c5aad9b650ad26ec412814/stpsf/utils.py#L208

Error (Python 3.10.9):
```python
TypeError: TarFile.extractall() got an unexpected keyword argument 'filter'
```

marcio-melendez:
This  is a small change to address issue #79 
By removing the keyword filter, then it uses the default value for the filter, None, which for versions of Python +3.10.12 it equivalent to 'fully_trusted' 
See 
https://docs.python.org/3.10/library/tarfile.html#extraction-filters

### DIFF
--- a/stpsf/utils.py
+++ b/stpsf/utils.py
@@ -205,7 +205,7 @@ def auto_download_stpsf_data():
 
             # Extract the tarball
             with tarfile.open(filename, "r:gz") as tar:
-                tar.extractall(default_path.parent, filter="fully_trusted")
+                tar.extractall(default_path.parent)
 
         if not any(default_path.iterdir()):
             raise IOError(f"Failed to get and extract STPSF data files to {default_path}")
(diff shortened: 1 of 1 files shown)


stpsf/utils.py
```diff
@@ -205,7 +205,7 @@ def auto_download_stpsf_data():
 
             # Extract the tarball
             with tarfile.open(filename, "r:gz") as tar:
-                tar.extractall(default_path.parent, filter="fully_trusted")
+                tar.extractall(default_path.parent)
```
BradleySappington, line 208: This is a correct fix, as the default (for now) is "fully_trusted".  I think it would be prudent to provide a small comment that `# TODO filter="fully_trusted" will need to be added as a parameter once python 3.14 is adopted due to advertised changing defaults in python.`

BradleySappington (changes requested):

stpsf/utils.py
marcio-melendez: Sounds good, I just added the note

BradleySappington (approved):
thanks Marcio!

PR #81, opened 2025-03-26
Roman: Update SIAF aperturename and pixel scale based on detector.

mperrin:
This small PR makes the Roman WFI implementation more consistent with the JWST instruments, by adding the same machinery for updating SIAF aperture name and pixel scale based on the selected detector.  The aperturename is inferred based on the detector, and the pixel scale is inferred as the average of the XSciScale and YSciScale values in the SIAF for that detector. 

(Currently it appears that the SIAF does not have distinct pixel scales for the WFI SCAs, but presumably it will eventually do so).

```
In [2]: wfi = stpsf.WFI()

In [3]: wfi.aperturename
Out[3]: 'WFI01_FULL'

In [4]: wfi.pixelscale
Out[4]: 0.1078577405

```

### DIFF
--- a/stpsf/roman.py
+++ b/stpsf/roman.py
@@ -267,6 +267,10 @@ class RomanInstrument(stpsf_core.SpaceTelescopeInstrument):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+
+        self.siaf = stpsf_core.get_siaf_with_caching('roman')
+        self._aperturename = None
+
         self.options['jitter'] = 'gaussian'
         self.options['jitter_sigma'] = 0.012  # arcsec/axis, see https://roman.ipac.caltech.edu/sims/Param_db.html#telescope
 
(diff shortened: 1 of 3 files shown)


marcio-melendez (approved):
Looks good.

braingram:
Is it expected that this would produce numerical differences for romancal usage? We're seeing some differences in nightly regression test runs:
https://github.com/spacetelescope/RegressionTests/actions/runs/14208621863
that so far I've been able to isolate to this PR. I can dig in further but I'm assuming this is caused by differences in the `SourceCatalogStep` in romancal which currently uses stpsf:
https://github.com/spacetelescope/romancal/blob/327f3f9d64350a2eb994e0937257a8f531c30040/romancal/source_catalog/psf.py#L121C11-L121C26
If the simulated psf differs I would expect the measured catalog to then differ (and finally the fit wcs to differ explaining the above test failures).

Thanks in advance for your help. Perhaps the current romancal code wasn't setting the pixel scale and with this PR it's being set automatically when the detector is assigned?

schlafly:
FYI, @braingram , I think this makes sense---the GriddedPSF objects we generate use the pixel scale to decide what size stamps to make, and so this small change means that those stamps will be (negligibly) different, which will lead to different PSF astrometry.  We could probably force the pixel scale to always be 0.11" or something, which IIRC was the old default.

mperrin:
Correct @schlafly, the old code defaulted to 0.11 always for all cases as the default. The new code is starting to try to make small steps to better (more precisely) represent or predict what will be seen on the detectors. 

You can easily set it back to `wfi.pixelscale = 0.11` if you want the old behavior for regression testing. But we should also collectively think about how you/we want to manage changes in PSF models in general for Roman. We are expecting a bunch of changes in the near future during continued efforts to iteratively improve the models for WFI.

braingram:
Thanks for the quick response and details. That's very helpful. It sounds like the current devdeps "failures" and improvements.

My understanding is that romancal will soon be switching to using psf reference files (like [this one](https://roman-crds.stsci.edu/browse/roman_wfi_epsf_0062.asdf) that were recently uploaded to crds. That may mean that romancal will no longer use stpsf directly and that future improvements will require regenerating new psf reference files and updating crds for romancal (and the regtests) to see those improvements.

Issue #85, opened 2025-05-04
`ValueError` about inhomogeneous `opds` shape in `show_wfs_during_program()`


vandalt:
Running `show_wfs_during_program` for program 2473 gives the error shown below. The issue is that some OPD files have shape `(128, 128)` instead of `(256, 256)`. I wrote a small code snippet to print OPD shape and the type of file. It seems that files with `OP_TYPE=MIMF` and `MODE=MANUAL` are the ones with shape 128. Most files have type `WAVEFRONT_MAINTENANCE`  and mode `AUTOMATIC` and shape `(256, 256)`. Several other files have type `FINE_PHASING` and mode `MANUAL`, but they also have shape `(256, 256)`.

As a quick fix, setting the start and end date to avoid problematic files works, but is not ideal.

Potential solutions would be to either modify the code to handle inhomogeneous shapes or to reject the `MIMF` files. Happy to contribute a PR with either of these, but I did not know enough about wavefront files to know which option is better.

Thanks!

**Code to reproduce**

```python
pid = "02473"
stpsf.trending.show_wfs_during_program(pid)
plt.show()
```

**Resulting error**

```python
Traceback (most recent call last):
  File "/home/vandal/scratch/stpsf_error_repro.py", line 4, in <module>
    stpsf.trending.show_wfs_during_program(pid)
  File "/home/vandal/miniforge3/envs/jyms/lib/python3.11/site-packages/stpsf/trending.py", line 1961, in show_wfs_during_program
    opd_array = np.asarray(opds)
                ^^^^^^^^^^^^^^^^
ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (301,) + inhomo
geneous part.
```

**Code to print info about OPD files**

```python
from astropy.io import fits
from pathlib import Path
stpsf_data_dir = Path(stpsf.utils.get_stpsf_data_path())
opd_files = stpsf_data_dir.glob("MAST_JWST_WSS_OPDs/*.fits")
for opd_file in opd_files:
    with fits.open(opd_file) as opd_hdul:
        opd_shape = opd_hdul[1].data.shape
        opd_hdr = opd_hdul[0].header
        if opd_shape == (128, 128):
            opd_file_128 = opd_file
        else:
            opd_file_256 = opd_file
        print(opd_file, opd_shape, opd_hdr["MODE"], opd_hdr["OP_TYPE"])
```
<details>

[...]

PR #87, opened 2025-05-06
Adding support for off-nominal phase map sizes 

marcio-melendez:
See #85 
This PR adds support for off-nominal phase map sizes, like those during commissioning. 
From this the trending function `show_wfs_during_program` can now show the WFE trending during commissioning programs, for example PID 1163 (initial wavefront maintenance)
`pid = "1163"`
`stpsf.trending.show_wfs_during_program(pid)`

image

It also solve the error as describe from the issue, in particular, it allows for a full display of PID 2473 without any issue by using the nominal date ranges. 
`pid = "02473"`
`stpsf.trending.show_wfs_during_program(pid)`

image

This PR also adds a hard cutoff day for the beginning of trending (the day of fine alignment, OTE-23). It also restric the padding not to be larger than 30 days. This avoid too much padding for programs contain within several JWST science cycles.

### DIFF
--- a/stpsf/mast_wss.py
+++ b/stpsf/mast_wss.py
@@ -800,6 +800,7 @@ def _query_program_visit_times_by_inst(program, instrument, verbose=False):
         'NIRCAM': 'Mast.Jwst.Filtered.NIRCam',
         'NIRSPEC': 'Mast.Jwst.Filtered.NIRSpec',
         'NIRISS': 'Mast.Jwst.Filtered.NIRISS',
+        'FGS': 'Mast.Jwst.Filtered.FGS'
     }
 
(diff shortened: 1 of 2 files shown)


mperrin (approved):
Overall looks good to me. Nice fixes! 

Minor comment: formally PEP8 says there should always be 2 spaces before a `#` comments symbol and one space after. (See https://peps.python.org/pep-0008/#inline-comments ) A few of the comments here don't follow that. This is a trivial edit though.

(And this reminded me that we didn't yet migrate the auto-tooling for PEP8 style review to this repo, which I posted about in Slack and will be addressed separately outside of the scope of this PR)

PR #92, opened 2025-06-06
assert len instead of return

BradleySappington:
Old broken test coming to light


marcio-melendez (approved):
This looks good. The test should not return a PSF obejct, so I think this check that the calculation was successful with a mask but not an image mask

PR #93, opened 2025-06-06
simplify test

BradleySappington:
simplify this test as we only care that calc_psf runs without errors


marcio-melendez (approved):
better. looks good.

PR #96, opened 2025-06-13
Fix display_psf_grid position labeling

### LINKED ISSUE #89: Minor display issue in `display_psf_grid`
I found a minor bug in `gridded_psf.display_psf_grid` where the grid positions print as `None` on the plot. It looks like these values used to be in a tuple, but that doesn’t seem to be the case anymore. 

Image

I will provide a quick fix that addresses the issues on both panels of this plot (only one panel shown here). The fix will be folded into an upcoming PR by @ojustino .

Image

ojustino:
@Skyhawk172 found that grid positions in `display_psf_grid` were being printed on the plot as `None` because of a type change and contributed a fix [in his display_grid_fix branch](https://github.com/Skyhawk172/stpsf/tree/display_grid_fix). I'm opening the PR with this branch instead to avoid that branch's initial merge commit.

Resolves #89.

### DIFF
--- a/stpsf/gridded_library.py
+++ b/stpsf/gridded_library.py
@@ -575,9 +575,8 @@ def display_psf_grid(grid, zoom_in=True, figsize=(14, 12), scale_range=1e-4, dif
     import matplotlib
     import matplotlib.pyplot as plt
 
-    def tuple_to_int(t):
-        if isinstance(t, tuple):
-            return (int(t[0]), int(t[1]))
+    def array_to_int(t):
+        return (int(t[0]), int(t[1]))
 
     def show_grid_helper(grid, data, title='Grid of PSFs', vmax=0, vmin=0, scale='log'):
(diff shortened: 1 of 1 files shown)


stpsf/gridded_library.py
```diff
@@ -575,9 +575,8 @@ def display_psf_grid(grid, zoom_in=True, figsize=(14, 12), scale_range=1e-4, dif
     import matplotlib
     import matplotlib.pyplot as plt
 
-    def tuple_to_int(t):
-        if isinstance(t, tuple):
-            return (int(t[0]), int(t[1]))
+    def array_to_int(t):
```
BradleySappington, line 578:
```suggestion
    def tuple_of_ints(t):
    # Convert first 2 elements of an array into a tuple of ints
```

BradleySappington (approved):

Issue #98, opened 2025-06-13
Reference issue in SUR.xmltext()


kulpster85:
I recently used the SUR.xmltext() function to generate SURs programmatically for JWST WFS&C analysis using TAS data and got a XML schema validation error with the resulting SUR.  I determined that the SUR attribute date/time values were missing the reference to self.

mperrin:
Closing. Fixed in #99

PR #99, opened 2025-06-13
#98 - Reference issue in SUR.xmltext()

kulpster85:
#98 

I recently used the SUR.xmltext() function to generate SURs programmatically for JWST WFS&C analysis using TAS data and got a XML schema validation error with the resulting SUR. I determined that the SUR attribute date/time values were missing the reference to self.

### DIFF
--- a/stpsf/opds.py
+++ b/stpsf/opds.py
@@ -1870,8 +1870,8 @@ class OTE_Linear_Model_WSS(OPD):
 
         """
 
-        if segment == 'SM':
-            raise ValueError('SM not supported by move_seg_local. Use move_sm_local instead.')
+        if segment == 'SM-19':
+            raise ValueError('SM-19 not supported by move_seg_local. Use move_sm_local instead.')
 
         # Handle tilts and clocking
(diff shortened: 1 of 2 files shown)


marcio-melendez (commented):
Hi Trey, 
This looks good, I just tested your changes and it is a good addition to the SUR xml. BTW, how did you get the initial  schema validation error? I am reading the sur via stpsf.surs.SUR(), then print a version of the SUR via sur.xmltext and then , read that back again using stpsf.surs.SUR(). If I do this I get an error but not a validation scheme one.

kulpster85:
The MCS throws the error I will send it to you since I don't think I should post here.

marcio-melendez (commented):
The functions_get_seg_sensitivities_from_sm and get_coeffs  will also need to change SM --> SM-19

marcio-melendez (approved):
This looks good, thanks for the changes

marcio-melendez:
> The functions_get_seg_sensitivities_from_sm and get_coeffs will also need to change SM --> SM-19

The SM in this case is related  to the OPD and not the SUR. so no need for the change

PR #100, opened 2025-06-17
Roman Cycle 10 Updates

### LINKED ISSUE #3: Change Roman detector naming from SCAxy to WFIxy

### LINKED ISSUE #4: Problem with Roman WFI PSFs and distortion
The problems, specifically, are that for SCA03, SCA06, SCA09, SCA12, SCA15, and SCA18, bluer monochromatic PSFs (around 0.8 microns and below) sum to values larger than 1, even when the flux is normalized to 1 at the pupil plane. The problem goes away when distortion effects are switched off.

This was discovered within the context of generating new PSFs for Pandeia, which does use a heavily customized script to create PSFs.

### LINKED ISSUE #63: Update Roman telescope+WFI optical models
In fall 2024, GSFC provided a new Roman optical model dataset for the telescope and WFI, including measured properties of the WFI optics as well as the measured figures of all the mirrors. 

This should be incorporated into STPSF.  Has been discussed internally in Slack; this is a GitHub issue to track also.

### LINKED ISSUE #65: High spatial frequency WFE maps for Roman OTA?
@AldenJurling, a quick question for you: Is there a publicly available file for the Roman OTE high spatial frequency WFE? 

Right now for the Roman WFI model in webbpsf, we have Zernike models for the low spatial frequencies, up to Z22. But no good information on the high spatial frequencies. As a proxy or placeholder, we use a version of the HST primary's high spatial frequencies, assuming that those are roughly representative.  But of course that's wrong in detail, a completely different telescope from different polishing technologies decades ago.

This would be good to improve on in some future release, so I'm curious if you can help provide any pointers or contacts for pursuing this. 

FYI @ojustino, @Skyhawk172

ojustino:
_(Resolves #3, resolves #4, resolves #63, resolves #65.)_

## `stpsf`

- [X] Where appropriate, changed mentions of `SCA` for detectors to `WFI` throughout the codebase and documentation.
<details>
  <summary>grep search (for posterity)</summary>

```
grep -ir  'SCA' PATH-TO-STPSF | grep -Ev '^.+stpsf.+:\s*.*\S{16,}(SCA|SCa|ScA|sCA|Sca|sCa|scA|sca)\S{16,}' | grep -Ev '^.+stpsf/[^/]+(\.csv|\.ipynb):' | grep -Ev '.ipynb_checkpoints|__pycache__|.git|.svg|.egg' | grep -Eiv 'scal|scat|scan|scar' | grep -i --color=always 'sca'
```

All case-insensitive instances of "SCA", except:
* Those followed and preceded by 16 word characters (letter/number/punctuation) [filters images in notebooks]
* Those in .csv or .ipynb files in the base STPSF repository [filters my test reference data and notebooks]
* Those in files with paths including .ipynb_checkpoints, etc. [filters files that aren't relevant for this purpose] 
* Those in which it is a substring of longer patterns like "scal," etc. [filters the many mentions of "scaling," "scatter," etc.]
</details>

### `RomanInstrument`
- [X] Locked `add_distortion` key of `RomanInstrument.options` dictionary, removed `add_distortion` keyword argument from `RomanInstrument.calc_psf()` (both of which are inherited by `WFI` class), and excluded Roman instruments from adding distortion in `gridded_library.CreatePSFLibrary` after determination that Zernikes are already given to us with distortion effects.

### `WFI`
- [X] Pointed `_aberration_files` attribute to Cycle 10's Zernike files.
  - [X] Remove `_get_filter_mode()` now that the list of "modes" (previously `imaging`, `prism`, `grism` in Cycle 9), or CSV files of Zernike coefficients, is now the same as the list of filters.
- [X] Changed `pupil_mask_list` attribute into a copy of the `filters` attribute.
  - The list of pupil masks (previously `SKINNY`, `WIDE`, `PRISM`, and `GRISM` in Cycle 9) is also now the same as the list of filters now that each filter now comes with its own pupil mask.
- [X] Modified `detector` setter to also update the `_filters` attribute with new file locations for filter throughput files. This is needed because in Cycle 10, unlike with other `stpsf` instruments, the WFI's per-filter throughput files are now also specific to each detector.
  - Also tried creating `RomanInstrument._get_filters()` for this purpose ([see in this commit](https://github.com/spacetelescope/stpsf/commit/8e28733502abcbce8a8ba2f546fbb848fdb0b810#diff-39c547afff718eb10d6901eeac0c9a2b793b38b3eb0b92918754c59181c70f21R378-R405)), but see dropdown below for rationale behind using the `detector` setter instead.

<details>
  <summary>how to adjust active throughput file</summary>

We could also supersede `SpaceTelescopeInstrument._get_filters()` with a `_get_filters()` in `RomanInstrument` or `WFI` that adds a "detector" to the filter_filename path, but we don't yet have `WFI._detector_list` (set in `WFI._load_detector_aberrations()` later in `WFI.__init__()`) or `WFI._detector` (set through `WFI.detector` later in `WFI.__init__()`). 

However...

1. We can't set `WFI._detector_list` earlier in `WFI.__init__()` because generating the path needed as an argument requires knowing the filters that the new `_get_filters()` method is trying to generate in the first place.
2. We can't set `WFI.detector` earlier in `WFI.__init__()` because `SpaceTelescopeInstrument.__init__()` will set `self._detector` back to None. Even a modification to `SpaceTelescopeInstrument.__init__()` like `if not hasattr(self, '_detector'): self._detector = None` (which I'm wary to do due to unintended effects on the JWST classes) wouldn't work because the `WFI.detector` setter triggers `WFI._update_pupil()`, which again requires knowledge of the filters.

We could go this route if we set `WFI._detector` (note underscore) instead, but doing so seems less clean than just using `detector` setter as we decided.

</details>

### `WFIPupilController`
- [X] Replaced ~`pupil_file_formatters`~ attribute with `pupil_file_formatters()` method to dynamically find a matching pupil file based on a given filter and detector.
- [X] Renamed ~`_get_filter_mask()`~ `_get_pupil_mask()` to increase clarity since the term "filter mask" is not used elsewhere. The method itself is now just returns the filter since, as noted in the **WFI** section, the list of filters is now the same as the list of pupil masks.
- [X] Moved validation of pupil mask from `lock_pupil_mask()` to `WFI.lock_pupil_mask()` since `WFIPupilController` lacks the direct access to the list of filters (which, again, is now identical to the list of pupil masks).

### Other
- [X] Regenerated figures in Roman documentation, the general README, and selected notebooks using the new optical models.
  - [x] still to go: remake `docs/roman_figures/pupil_mask_by_detector.gif`
- [X] Updated Roman documentation on pupil masks and added notes about distortion in Roman docs and `Distortion_examples.ipynb` notebook.
- [X] Updated tests

### New data files

(To examine file processing procedures, see `stpsf-data-source/WFI/sources/cycle10/Process_zern_and_filters_cycle_10.ipynb` for Zernikes and filter files or `webbpsf-data-source/WFI/sources/cycle10/Oversample_rescale_pupil_images_cycle_10.ipynb` for pupil files. Also added new notebook `stpsf-data-source/WFI/sources/cycle10/Create_pupil_gif.ipynb` with instructions on how to create the pupil mask GIF seen in the Roman documentation.)

- [x] **"Pre-archived"** Cycle 10 data files in `stpsf-data-source/WFI/sources/cycle10/` and then copied them to their appropriate locations in `stpsf-data-source` (for `upscaled_HST_OPD.fits`) and `stpsf-data-source/WFI/` (for the rest).
- [x] **Updated** Zernike files
  - imaging filters: `stpsf-data-source/WFI/sources/cycle10/WIM_FNNN_zernikes_cycle10.csv` and `stpsf-data-source/WFI/sources/cycle10/WIM_W148_zernikes_cycle10.csv`
  - spectral filters: `stpsf-data-source/WFI/sources/cycle10/PRISM_W148_zernikes_cycle10.csv` and `stpsf-data-source/WFI/sources/cycle10/WSM_GRISMN_zernikes_cycle10.csv`
- [x] **Updated** filter bandpass file (`stpsf-data-source/WFI/sources/cycle10/filters.tsv`)
- [x] **Created** filter throughput files,  now specific to each filter *and* each detector  (`stpsf-data-source/WFI/sources/cycle10/filters/WFINN/*_throughput.fits`).
- [x] **Updated** OPD file to match new pupil images' pixel scale (`stpsf-data-source/upscaled_HST_OPD.fits`).


docs/roman.rst
```diff
@@ -12,35 +15,59 @@ Wide Field Instrument (WFI)
...
    :alt: Sample PSFs for the filters in the Roman WFI.
 
-   Sample PSFs for the filters in the Roman WFI on SCA01. Angular
+   Sample PSFs for the filters in the Roman WFI on detector WFI01. Angular
    scale in arcseconds, log-scaled intensity. Note that the prism and
    grism PSFs shown here are monochromatic.
 
-The WFI model is based on the `Cycle 9 instrument reference information <https://roman.gsfc.nasa.gov/science/Roman_Reference_Information.html>`_ from the Roman team at Goddard Space Flight Center (GSFC). The reported jitter for the Roman observatory is 0.012 arcsec per axis, per `GSFC <https://roman.ipac.caltech.edu/sims/Param_db.html#telescope>`_.
+The WFI model is based on the
+`Cycle 10 instrument reference information <https://roman.gsfc.nasa.gov/science/Roman_Reference_Information.html>`_
```
ojustino: It looks like GSFC removed the spreadsheet at this link. (They previously had a copy of the spreadsheet we used during the Cycle 9 update here.) Should we keep the link? Go somewhere else?

docs/roman.rst
```diff
@@ -12,35 +15,59 @@ Wide Field Instrument (WFI)
...
+assess the impact of inter-detector and intra-detector PSF variations on
+science cases of interest.
 
 .. note::
 
    *Tutorial notebook for Roman*
 
-   This documentation is complemented by an `IPython Notebook tutorial for Roman PSFs <http://nbviewer.ipython.org/github/spacetelescope/stpsf/blob/stable/notebooks/STPSF-Roman_Tutorial.ipynb>`_. Download and run that notebook to use a beta notebook GUI for the WFI model, and to explore code samples for common tasks interactively.
+   This documentation is complemented by an
+   `IPython Notebook tutorial for Roman PSFs <http://nbviewer.ipython.org/github/spacetelescope/stpsf/blob/stable/notebooks/STPSF-Roman_Tutorial.ipynb>`_.
```
ojustino: This link leads to a 404 page. Is STPSF still working on NBViewer? If not, should we just link to the notebook's GitHub page? Or somewhere on the Roman Research Nexus?

ojustino (commented):
A couple of questions about links in the Roman documentation.

docs/roman.rst
marcio-melendez: I think we should move STPSF-Roman_Tutorial.ipynb to the documentation and link to the readthedoc page.

stpsf/roman.py
```diff
@@ -523,41 +554,68 @@ def set_base_path(self, datapath):
...
+        detector : string
+            See WFI.detector_list for a list of valid detectors.
+        """
+        if wfi_filter.upper().startswith(('F', 'W')):
+            # return f"RST_WIM_Filter_{wfi_filter}_{detector[:3]}_{str(int(detector[3:]))}.fits" #.gz? old
+            return f"RST_WIM_Filter_{wfi_filter}_{detector}.fits.gz"
+        elif wfi_filter.upper().startswith('GRISM'):
+            # STARTSWITH NOT NEEDED IF GRISM IS ONLY ONE ORDER NOW
+            # return f"RST_WSM_GRISM_grism_{detector[:3]}_{str(int(detector[3:]))}.fits" #.gz? old
+            return f"RST_WSM_GRISM_grism_{detector}.fits.gz"
```
Skyhawk172: The filenames in the new STPSF data folder do not have the GRISM and PRISM strings capitalized. As a result, the code is not finding the pupil masks when simulating a Grism or Prism PSF. This can be fixed with using "RST_WSM_Grism_grism_{detector}.fits.gz" and "RST_WSM_Prism_prism_{detector}.fits.gz"

stpsf/roman.py
```diff
@@ -779,6 +858,63 @@ def aperturename(self, value):
...
+    # `if not hasattr(self, '_detector'): self._detector = None`
+    # (which I'm wary to do due to unintended effects on the JWST classes)
+    # because WFI.detector's setter triggers WFI._update_pupil(), which
+    # again requires knowledge of the filters. COULD do this if we set
+    # WFI._detector (note underscore) instead, but do we want to skip the
+    # setter in that scenario?
+    ################################
+
     @RomanInstrument.filter.setter
     def filter(self, value):
```
Skyhawk172, line 824: The string for F146 now shows as W146 in filter_list. The filename for the throughput file in the new STPSF data also shows W146. This should be changed back to F146.

docs/roman.rst
Skyhawk172: They removed the file from the website a while ago and it's not available anywhere other than on request, so let's remove this particular link. We might want to ask about posting it to the Goddard Github technical information repo: https://github.com/spacetelescope/roman-technical-information/tree/main

stpsf/roman.py
ojustino: This is a good catch – thanks. It's updated in the newest commit.

stpsf/roman.py
ojustino: The filter is relabeled as W146 across the spreadsheet, pupil images, etc. delivered from GSFC. I hadn't seen W146 before, but it seemed like a lot of occurrences to be a typo. Do you know where it might come from?

Skyhawk172: I think this is a relic of earlier notation for the different filters... All of our current documentation, including Goddard's, refers to this filter as F146.  APT and the ETC also use F146.

Skyhawk172:
### Testing of Cycle 10 Roman Optical Model

Click to expand

<details>
<summary>Cycle 10 vs Cycle 9 PSF on WFI10</summary>

animated_imaging_WFI10

animated_spectro_WFI10
</details>

<details>
<summary>Cycle 10 PSFs loop all detectors</summary>

animated_imaging_cycle10

animated_spectro_cycle10
</details>


<details>
<summary>Cycle 10 vs Cycle 9 Zernikes</summary>

animated_zernikes_wfi062um

animated_zernikes_prism120um

</details>

I have also separately checked the following:
- [x] Disabling of SIAF distortion
   - There is no more `add_distortion` argument to the roman `calc_psf()` function.
   - Directly setting `wfi.options["add_distortion"] = True` is disabled; a warning is also output "WFI: add_distortion disabled for Roman PSFs. Calculating DET_SAMP extension"
   - We maintain the four extensions in the final output FITS HDU for backward compatibility; however, I confirm that the two oversampled extensions are identical. Same for the two detector-sampled extensions.

- [x] Input Zernike spreadsheets
   - The values in the individual Zernike spreadsheets match those from the original delivery, including the Grism 0-0 and 1-1 orders (only did spot checks).
   - Also, the Zernike plots above seem to match the values from these spreadsheets (again only did spot checks).

- [x] Selection of appropriate input files
   - I checked that the correct Zernike and pupil files were being retrieved when changing the filter and/or detector (using `wfi._current_aberration_file` `wfi.pupil`).

- [x] Compare PSF with Goddard's PSF
    - see [comment below](https://github.com/spacetelescope/stpsf/pull/100#issuecomment-2985496837)

- [x] SCA-->WFI detector string
    - Detector are now referred to as `WFIxy`. 
    - However, for backward compatibility, users can still use `SCAxy` and STPSF will change it to `WFIxy`.

stpsf/roman.py
Skyhawk172: The file `WIM_W146_zernikes_cycle10.csv` will also need to be renamed. Same for the pupil mask filenames.

notebooks/Distortion_examples.ipynb
```diff
@@ -62,9 +62,10 @@
     "* MIRI PSFs have the distortion based on the detector shape and a scattering effect from the MIRI detector applied\n",
     "    * Note that MIRI LRS mode is not currently supported\n",
     "\n",
-    "\n",
     "* NIRSpec PSFs have just the distortion based on the detector shape applied\n",
     "\n",
+    "* As of STPSF 2.1, the Roman WFI's Zernike coefficients are evaluated from a computed wavefront map with all apertures removed from the optical model. They are defined on a 90 mm circle at the exit pupil. As such, they are already distorted, so **Roman WFI PSFs are distorted by default.** This distortion cannot be removed.\n",
```
Skyhawk172: Actually, here, I think the distortion contribution is accounted for in the pupil masks, and not in the Zernikes.

stpsf/distortion.py
```diff
@@ -125,7 +125,7 @@ def distort_image(
         instrument = hdu_list[0].header['INSTRUME'].upper().strip()
 
         if instrument == 'WFI':
-            aper_name = 'WFI' + hdu_list[0].header['DETECTOR'][-2:] + '_FULL'
+            aper_name = hdu_list[0].header['DETECTOR'] + '_FULL'
```
Skyhawk172, line 128: These changes look good, even though we won't be using these functions for now, or at least until we get undistorted pupil masks from Goddard.

mperrin:
A couple big picture comments/questions

1. How much do you want to prioritize back compatibility, not breaking existing code? For instance if changing the names of elements or what’s in the pupil_list etc, do you care if that breaks existing code? For the JWST side we have in general tried to avoid breaking existing code when possible, for instance by adding checks so the older values are still allowed as inputs and get converted into the appropriate new values. 

2. I have a particular concern re dropping the additional extensions. Recall that these in principle include both the optical distortion and ALSO the detector effects (IPC etc). I’ve emphasized before how the detector charge transfer for JWST was the single biggest piece we missed before flight and it seems worth trying to include for Roman. If you take out the additional extensions now, it (a) breaks any existing code that expects there to be 4 extensions, and (b) makes it harder in the future to encourage folks to use an extension that includes the charge transfer effects. 

Think about instead an approach where we don’t remove those extensions, we just have the Roman code path for distortion skipped lower down so there’s no change in the pixel values. I think that might be better for multiple reasons, including increasing the consistency between the Roman and JWST output data products. (All will have 4 extensions, and we can recommend using DET_DIST extension 3 in all cases as the one that best matches science frame 2d calibrated data, for examples)

stpsf/roman.py
```diff
@@ -216,11 +218,12 @@ def _load_wfi_detector_aberrations(filename):
 
     def build_detector_from_table(number, zernike_table):
         """Build a FieldDependentAberration optic for a detector using
-        Zernikes Z1-Z22 at various wavelengths and field points"""
+        Zernikes Z1-Z43 at various wavelengths and field points"""
```
Skyhawk172: The original spreadsheet goes all the way up to Z45.

stpsf/roman.py
```diff
@@ -251,11 +254,25 @@ def build_wavelength_dependence(rows):
...
+        detectors['WFI{:02}'.format(detid)] = build_detector_from_table(detid, zernike_table)
 
     return detectors
 
 
+class _RomanInstrumentOptionsDict(dict):
+    """
+    A wrapper for the options dict that prevents the `add_distortion` key from
+    being set to True or False since distortion can't be added to PSFs from
+    Roman WFI (they come pre-distorted) or the Coronagraph Instrument
```
Skyhawk172: I'd say "input data comes already distorted" as opposed to "pre-distorted".

stpsf/roman.py
```diff
@@ -271,8 +288,13 @@ def __init__(self, *args, **kwargs):
...
         self._aperturename = None
 
+        # reassign self.options from dict to _RomanInstrumentOptionsDict,
+        # mainly to prevent user modification of the add_distortion key
+        self.options['add_distortion'] = 'NA'  # distortion is always on
+        self.options = _RomanInstrumentOptionsDict(self.options.copy())
+
         self.options['jitter'] = 'gaussian'
-        self.options['jitter_sigma'] = 0.012  # arcsec/axis, see https://roman.ipac.caltech.edu/sims/Param_db.html#telescope
+        self.options['jitter_sigma'] = 0.012  # arcsec/axis, see https://roman.ipac.caltech.edu/page/param-db#telescope
```
Skyhawk172: Let's point to the GSFC Roman Technical Information repo instead: https://github.com/RomanSpaceTelescope/roman-technical-information/tree/main/data/Observatory/MissionandObservatoryTechnicalOverview

stpsf/roman.py
ojustino: For completeness, this also involves renaming the filter's entry in `filters.tsv`. I have reverted to the original filter name in all of these places and copied the updated reference data to the same location as earlier.

stpsf/roman.py
```diff
@@ -633,31 +692,41 @@ def __init__(self):
...
+        #     'imaging': os.path.join(self._datapath, 'wim_zernikes_cycle9.csv'),
+        #     'prism': os.path.join(self._datapath, 'wsm_prism_zernikes_cycle9.csv'),
+        #     'grism': os.path.join(self._datapath, 'wsm_grism_zernikes_cycle9.csv'),
+        #     'custom': None,
+        # }
+        self._aberration_files = {'custom': None}
+        self._aberration_files.update({
+            fltr: os.path.join(
+                self._datapath,
+                (f"{'WIM' if fltr.startswith(('F', 'W')) else 'WSM'}"
```
Skyhawk172: Now that W146 was renamed to F146, we can remove the `starstwith("W")` part of this line.

Skyhawk172 (commented):
Minor comments throughout.

ojustino:
Thanks for taking a look, @mperrin. 

**On backward compatibility,** we do prioritize it with most user-facing functionality. For example, these changes still allow users to refer to detector "SCA01" and have that converted to "WFI01" behind the scenes. Anything related to `WFIPupilController` is considered internal (since users would access it through `wfi._pupil_controller`), so we have more leeway there.

The `pupil_mask_list` is user-facing, but it has already tended to change with each optical model update – see [the table at the end of the pupil mask section](https://stpsf.readthedocs.io/en/v2.0.0/roman.html#pupil-variation-and-pupil-masks-in-the-wfi-model) of the current Roman STPSF documentation for past iterations. Now that we have one pupil mask per filter, I think the list should stabilize going forward.

**On additional extensions,** I agree with your concerns and will work on implementing your suggestion before we merge anything.

mperrin:
👍 @ojustino . 

As one specific suggestion, I think you could look at just commenting out [this one line](https://github.com/spacetelescope/stpsf/blob/422de4cbd85784fbf3d777e9a024d06f2925586a/stpsf/roman.py#L418) in `roman.py`, the one with the call to `apply_distortion`. Doing so should result in the creation of the extensions with copies of the first two extensions.  The subsequent line you'll need to tweak a variable name in the code but that would apply the detector charge transfer model.

notebooks/Distortion_examples.ipynb
ojustino: Thanks for these finds. I attempted to address this and the other comments in my latest commit.

Skyhawk172:
**[UPDATED]** Regarding comparing with the PSF that GSFC sent us, here are some examples. I had initially reported some differences between these PSFs and the ones from STPSF, but we tracked that down to the number of wavelengths to use; in other words, the GSFC PSF are monochromatic whereas my initial tests were not.  

Here are the comparisons for the F158, Prism, and Grism elements. Note that I also set the jitter to 0. I still haven't gotten  to match the pixel sizes between the simulated images to do a difference image, but the agreement is qualitatively very good.

<details>
<summary>Updated F158 monochromatic</summary>

Cycle10-STPSF-vs-GSFC-PSF_F158
</details>

<details>
<summary>Updated Prism monochromatic</summary>

Cycle10-STPSF-vs-GSFC-PSF_PRISM
</details>

<details>
<summary>Updated Grism1 monochromatic</summary>

Cycle10-STPSF-vs-GSFC-PSF_GRISM1
</details>

<details>
<summary>Initial test (polychromatic sims)</summary>

This is the initial test that prompted us to look more closely into it, until we soon realized the GSFC PSFs are in fact monochromatic.
Cycle10-STPSF-vs-GSFC-PSF
</details>

marcio-melendez (commented):
There are lots of comments in this PR and I am not sure if I am reviewing the latest commit. In any case, some of these initial comments may be repeats, mostly on file name format but that give errors when running a calculation. 
1. Currently under the new Cycle 10/WFI/pupil , the filter file for F146 is RST_WIM_Filter_W146_WFI01.fits.gz but it should be RST_WIM_Filter_F146_WFI14.fits.gz
2. Both prism and grism expect files of the type  RST_WSM_PRISM_prism and RST_WSM_GRISM_grism, but in stpsf-data-cyc10/WFI/pupils the names are lower case, RST_WSM_Prism_prism and RST_WSM_Grism_grism

ojustino:
Hi @obi-wan76, thank you for your comments.
1. Please check the pupils at my reference data path again; the filter names should be what you expect.
2. I addressed that [earlier today](https://github.com/spacetelescope/stpsf/pull/100#discussion_r2153393832), so you may need to update your branch. The latest commit at the moment is [484db0e](https://github.com/spacetelescope/stpsf/pull/100/commits/484db0e219e28a2aaad5627ae913bba16a5137cd).

marcio-melendez:
> Hi @obi-wan76, thank you for your comments.
> 
> 1. Please check the pupils at my reference data path again; the filter names should be what you expect.
> 2. I addressed that [earlier today](https://github.com/spacetelescope/stpsf/pull/100#discussion_r2153393832), so you may need to update your branch. The latest commit at the moment is [484db0e](https://github.com/spacetelescope/stpsf/pull/100/commits/484db0e219e28a2aaad5627ae913bba16a5137cd).

Thanks! yes, I had an old commit.

Skyhawk172:
I have updated my [comment above](https://github.com/spacetelescope/stpsf/pull/100#issuecomment-2985496837) regarding comparisons with GSFC PSF with new figures showing _monochromatic_ PSFs for F158, Prism, and Grism. You'll see that the match is qualitatively much better now.

docs/roman.rst
ojustino: So it's recorded here, our preference is not to duplicate the notebook onto another webpage after [already having done so in RDox](https://roman-docs.stsci.edu/simulation-tools-handbook-home/stpsf-for-roman/stpsf-tutorials). I'm still not sure about the NBViewer link, but it seems like linking to the notebook's GitHub page might be best for now.

mperrin: I think the NBviewer link may just be an old remnant from back years ago before Github natively supported rendering notebooks

ojustino:
I just pushed a rebased version of the branch that, as discussed above, reverts the removal of most of `RomanInstrument._calc_psf_format_output()` so we continue to create four extensions per PSF. I also folded in updates to the links in my initial review and a new version of the pupil image GIF from the Roman docs page.

Tests are passing for me locally, so hopefully that will be reflected here once we get the new reference files on Box. I have some comments in `stpsf/roman.py` that I'll take out once we have agreement that the restructuring is good to go. Other than that, I think we're very close since the PSFs are looking all right.

stpsf/roman.py
```diff
@@ -682,10 +771,10 @@ def _load_detector_aberrations(self, path):
         path : string
             The path to the file containing detector aberrations.
         """
-        detectors = _load_wfi_detector_aberrations(path)
+        detectors_dict = _load_wfi_detector_aberrations(path)
         assert len(detectors.keys()) > 0
```
Skyhawk172: I'm getting an error here.

This line should be `assert len(detectors_dict.keys()) > 0` instead

ojustino: Just changed in the new commit. I was prompted to make this change after noticing a name clash with the `detectors` imported at the top of the file.

Skyhawk172:
I'm testing the handling of additional extensions (i.e. ext 2 and 3) now that we are not adding any distortion. 

I confirm that we do indeed get four extensions now, and that extensions 0 and 2 (oversampled) are identical; same for extension 1 and 3 (detector-sampled).

The function `detector.apply_detector_charge_diffusion()` get called for Roman WFI even if the `charge_diffusion_sigma` is zero for Roman, which leads to some log messages and headers that can be misleading. I propose that we skip some logging and header keywords if that option is zero:
     
```
psf_hdulist[ext].data = psf_hdulist[ext].data
if sigma > 0.:
    out = scipy.ndimage.gaussian_filter(psf_hdulist[ext].data, sigma / psf_hdulist[0].header['PIXELSCL'])
    psf_hdulist[ext].data = out
    stpsf.stpsf_core._log.info(
    'Detector charge diffusion: Convolving with Gaussian with sigma={0:.3f} arcsec'.format(sigma))
    psf_hdulist[ext].header.add_history('Applied detector charge diffusion model.')
    psf_hdulist[ext].header['CHDFTYPE'] = ('gaussian', 'Type of detector charge diffusion model')
    psf_hdulist[ext].header['CHDFSIGM'] = (sigma, '[arcsec] Gaussian sigma for charge diff model')
```

ojustino:
@Skyhawk172 I believe this is similar behavior to past versions since `RomanInstrument._calc_psf_format_output()` only checked whether `add_distortion` was True before calling `stpsf.detectors.apply_detector_charge_diffusion()`, even if `charge_diffusion_sigma` was zero. But I see what you mean about making the logging  more accurate.

I tried a slightly different approach that removes all charge-diffusion related logging from `RomanInstrument._calc_psf_format_output()` and returns `psf_hdulist` as-is in `stpsf.detectors.apply_detector_charge_diffusion()` if `charge_diffusion_sigma` is zero while also logging a dedicated comment for that scenario. Let me know if you support this approach.

stpsf/detectors.py
```diff
@@ -233,6 +233,11 @@ def apply_detector_charge_diffusion(psf_hdulist, options):
 
     sigma = options.get('charge_diffusion_sigma')
 
+    if sigma == 0:
```
Skyhawk172: In the case of Roman (at least), `sigma = None` when we enter this method, so the `if sigma == 0` check should go _after_ the `if sigma is None` check. This way, the latter will set `sigma = 0` and then we can return out of the method.

ojustino: All right, please see the latest commit. In `stpsf/constants.py` the WFI is the only instrument with zero charge diffusion by default, at least for now.

Skyhawk172:
@ojustino This PR looks very good to me now. 

One more question: do you think we can resolve [#4](https://github.com/spacetelescope/stpsf/issues/4) as part of this PR? I just ran some checks and the normalization appears to be working as expected now that we've removed the SIAF distortion.

ojustino:
@Skyhawk172 Sure, since `add_distortion` seemed to be the issue, your improved results make sense to me. I'll add it to the list of resolved issues in the top comment.

Skyhawk172:
@ojustino Good catch on the grids. Thinking about backward compatibility again, I think we should allow users to specify `add_distortion = True` when calling `psf_grid()` for Roman, but then disable it along with the warning message, just like it's done with `calc_psf()`, i.e. "WFI: add_distortion disabled for Roman PSFs". This way, their "old" code will still work.

ojustino:
@Skyhawk172 The warning you're referencing is different. That's for setting `wfi.options['add_distortion']` to True or False. The current result for `wfi.calc_psf(add_distortion=True)` is a `TypeError` referencing `add_distortion` as an unexpected keyword argument, and that's what happens with `wfi.psf_grid(add_distortion=True)` now as well.

I'm concerned that allowing users to specify `add_distortion` in either function and successfully obtain a result may give those who don't process the logging messages the impression that it's still possible to toggle distortion. Perhaps the error could give more explicit instructions on its resolution (just remove `add_distortion` as an argument). What do you think?

Skyhawk172:
@ojustino My point was simply to tell the user that distortion is disabled for Roman grids (just like it is for single PSF calculations). So "older" code that included distortion into grids would have to be updated, but that's a minor change.

ojustino:
@Skyhawk172 You mentioned including a "warning message, just like it's done with `calc_psf()`," but `wfi.calc_psf()` doesn't log any messages about distortion after the changes you requested to the charge diffusion code. The only distortion-related warning currently logged in the Roman code is if a user tries to assign a value to `wfi.options['add_distortion']`, separate from PSF creation. 

You also said "we should allow users to specify `add_distortion = True` when calling `psf_grid()`," but that is also different from what is currently implemented with `wfi.calc_psf()`. Both `wfi.calc_psf()` and `wfi.psf_grid()` currently rely on documentation and errors to steer users away from the `add_distortion` argument.

I understand you as instead saying you'd like to allow use of the `add_distortion` argument but log a warning that it has no effect on the PSFs. But that allows someone who uses the argument and misses the logs to think they've added/removed distortion when they really haven't. I prefer the current approach because it's more explicit about what isn't allowed.

Skyhawk172:
@ojustino and I chatted offline and are all on the same page regarding the above comments on the `psf_grid()` warning.

mperrin:
Hmm, something else to keep in mind about having the add_distortion raise a TypeError for wfi.calc_psf, is that makes the API of the Roman and JWST versions inconsistent. I know, the behaviors are going to be different in any case, but is there an argument for API consistency?

edit: but if you both are 👍 then OK with me!

Skyhawk172 (approved):
I've fetched the latest commits from @ojustino, pointed my STPSF_PATH to the repo containing all the new changes/additions, and successfully re-ran all of my tests. As far as I'm concerned, this PR is good to go.

ojustino:
@mperrin Our eventual approach was to accept the `add_distortion` argument with `None` as the default value and trigger a `ValueError` with a more descriptive message if it's assigned a different value. If the error is not triggered, we just skip assigning an `add_distortion` key to the `options` dictionary like in earlier commits in this PR. This should keep the API consistent and hopefully give more helpful advice to those using code that's now outdated.

PR #102, opened 2025-06-25
V2.1.0

BradleySappington:
Release V2.1.0


braingram:
Can this be delayed until https://github.com/spacetelescope/stpsf/issues/103 is addressed? Making this release will break released versions of romancal. Let me know if more information is helpful.

Issue #103, opened 2025-06-25
Allow `add_distoration=False` for roman


braingram:
One of the changes in https://github.com/spacetelescope/stpsf/pull/100 is causing failures in romancal testing (and will cause failures for released romancal when stpsf is released).

Romancal currently provides `add_distoration=False` to `calc_psf`:
https://github.com/spacetelescope/romancal/blob/d0fc728e41f058785858e05b891b35207199dc7c/romancal/source_catalog/psf.py#L255

on stpsf develop this produces a `TypeError` and causes test failures. Can this code be updated (prior to the next release) to not produce an error if `add_distortion=False`?

We can update romancal main to remove `add_distortion` but released versions of romancal will fail with the stpsf develop.

braingram:
xref: https://github.com/spacetelescope/romancal/pull/1813 for work trying to make romancal compatible with the recent stpsf changes.

mperrin:
FYI @ojustino @Skyhawk172 @BradleySappington

BradleySappington:
@mperrin - have all already been talking via call, thanks for the heads up

braingram:
Fixed by: #104

PR #104, opened 2025-06-25
Re-enable add_distortion argument for Roman

ojustino:
We were asked in #103 to hold off on the decision in #100 to trigger an error if the `add_distortion` argument of `wfi.calc_psf()` (and `wfi.psf_grid()`) is not `None`. While this PR does not change the background behavior that keeps this argument from affecting simulated Roman PSFs, it now triggers warnings instead of errors.

In the future, I think it would still be best to trigger an error in this situation. This PR is a compromise for now since our upstream dependencies need more time to react to the change and we also have some time pressure on the v2.1.0 release.


Skyhawk172 (approved):
I've tested this PR with the `add_distortion` keywords to both `calc_psf` and `psf_grid` and the behavior is as expected. The functions now accept the argument, but it has no effect on the final PSF. Warnings and docstring were updated to reflect the fact that this argument is for backward compatibility only. This applies to Roman WFI only.

PR #113, opened 2025-08-18
Update ci with prerelease

BradleySappington:
Update the CI to use the new PRE-RELEASE data on box.
When data changes we want to be able to still verify using our CI without altering the data other repositories use.

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 jobs:
   retrieve_cache:
-    uses: spacetelescope/stpsf/.github/workflows/retrieve_cache.yml@1b4adc279d0e3c76e01d975fb2a46a10258b5d53 # setup retrieve_cache with development or latest_release
+    uses: spacetelescope/stpsf/.github/workflows/retrieve_cache.yml@00753c8ce7ca592757825061d3ee2fa36ec329df # setup retrieve_cache with development or latest_release
     with:
       dataset_type: development   # or latest_release
   tests:
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -4,7 +4,9 @@ on: [push, pull_request]
 
 jobs:
   retrieve_cache:
-    uses: spacetelescope/stpsf/.github/workflows/retrieve_cache.yml@68bfa1fe978b05c892672e715f0666ecf999d32f  # v2.1.1
+    uses: spacetelescope/stpsf/.github/workflows/retrieve_cache.yml@68bfa1fe978b05c892672e715f0666ecf999d32f  # v2.1.1 - prerelease implementation
+    with:
+      dataset_type: prerel   # or "full"
   tests:
     needs: [ retrieve_cache ]
     name: ${{ matrix.name }}
@@ -13,7 +15,6 @@ jobs:
       fail-fast: false
       matrix:
         include:
-
           - name: Coverage test in Python 3
             os: ubuntu-latest
             python: '3.11'
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 jobs:
   retrieve_cache:
-    uses: spacetelescope/stpsf/.github/workflows/retrieve_cache.yml@main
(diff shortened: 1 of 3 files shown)


.github/workflows/retrieve_cache.yml
```diff
@@ -27,43 +27,53 @@
...
+          # Path where cache would be restored
+          echo "cache_path=/tmp/stpsf-cache/$CACHE_KEY" >> $GITHUB_OUTPUT
+
+      - name: Set outputs
+        id: set-outputs
+        run: |
+          echo "cache_key=${{ steps.cache-key.outputs.cache_key }}" >> $GITHUB_OUTPUT
+          echo "version=${{ steps.cache-key.outputs.version }}" >> $GITHUB_OUTPUT
+          echo "cache_path=${{ steps.cache-key.outputs.cache_path }}" >> $GITHUB_OUTPUT
+
```
zacharyburnett, line 75:
```suggestion

    outputs:
      cache_path: ${{ steps.set-outputs.outputs.cache_path }}
      cache_key: ${{ steps.set-outputs.outputs.cache_key }}
      version: ${{ steps.set-outputs.outputs.version }}
```

.github/workflows/retrieve_cache.yml
```diff
@@ -27,43 +27,53 @@
...
 
 jobs:
-  retrieve_latest_cache_key:
-    name: retrieve latest STPSF data cache key
+  retrieve:
     runs-on: ubuntu-latest
+    outputs:
+      cache_path: ${{ steps.set-outputs.outputs.cache_path }}
+      cache_key: ${{ steps.set-outputs.outputs.cache_key }}
+      version: ${{ steps.set-outputs.outputs.version }}
```
zacharyburnett:
```suggestion
```

.github/workflows/ci_workflows.yml
```diff
@@ -4,9 +4,9 @@ on: [push, pull_request]
 
 jobs:
   retrieve_cache:
-    uses: spacetelescope/stpsf/.github/workflows/retrieve_cache.yml@098e3d3515d906bbfa12680d2282ec8d6726527b  # v1.5.2
+    uses: spacetelescope/stpsf/.github/workflows/retrieve_cache.yml@1b4adc279d0e3c76e01d975fb2a46a10258b5d53 # setup retrieve_cache with development or latest_release
```
zacharyburnett: make sure this commit hash is available on `develop` after this branch is merged!

.github/workflows/retrieve_cache.yml
zacharyburnett: moves `outputs` to the end for clarity

.github/workflows/retrieve_cache.yml
zacharyburnett: moves `outputs` to the end for clarity (see below)

.github/workflows/retrieve_cache.yml
```diff
@@ -27,43 +27,53 @@
...
-      cache_key: ${{ steps.latest_cache_key.outputs.cache_key }}
+          PREFIX="stpsf-data-${{ inputs.dataset_type }}-"
+
+          # Find the most recent cache key matching dataset type
+          CACHE_KEY=$(gh actions-cache list -R ${{ github.repository }} --key "$PREFIX" --sort created-at | cut -f1 | head -n1)
+
+          echo "cache_key=$CACHE_KEY" >> $GITHUB_OUTPUT
+
+          # Extract version (suffix after prefix, if present)
+          VERSION=${CACHE_KEY#"$PREFIX"}
```
zacharyburnett, line 63: how does this work? I don't think I've seen this syntax before

zacharyburnett:
This looks like it should work, as far as I can tell. Good to see an ability to select a pre-release!

.github/workflows/retrieve_cache.yml
BradleySappington: this is a bash syntax that will strip $PREFIX (defined above) from the CACHE_KEY leaving only the version which is saved as VERSION.  You can try it in terminal :) 

`$ PREFIX="stpsf-data-development-"
$ TEMP_KEY="stpsf-data-development-20250818"
$ echo ${TEMP_KEY#"$PREFIX"}`

.github/workflows/ci_workflows.yml
BradleySappington: Hash's should be retained through commit history so should be good to go

PR #114, opened 2025-08-19
Implement model for IEC-driven WFE astigmatism

mperrin:
This PR implements a model for the IEC-driven WFE, as described in Telfer et al. 2024. This includes: 

 - Retrieval from MAST of the relevant IEC heater telemetry during a specified time range
 - Calculation of the IEC astigmatism coefficient, following the method described in Randal's paper. (using a slightly updated set of coefficients he provided, more recent than that paper)
 - Using that coefficient and the known pattern of segment PTTs characteristic of this mode to update the OTE WFE model. 

This model replaces the prior pre-flight IE model.

Example code:
```
    time = '2025-08-01T12:00:00'  
    iec_coeff = stpsf.opds.estimate_iec_induced_wfe_at_time(time)

    nrc = stpsf.NIRCam()
    nrc, ote = stpsf.enable_adjustable_ote(nrc)
    
    ote.apply_iec_drift(iec_coeff)

```

The `estimate_iec_induced_wfe_at_time` function will work on vector inputs for `time` to return a vector of IEC WFE amplitude over time, for cases where one wants to compute a time series of PSFs with this small variation.  

That function also has a `plot=True` keyword, to display how it's working behind the scenes: 
```
stpsf.opds.estimate_iec_induced_wfe_at_time('2025-08-18T00:00:00', plot=True)
```
Screenshot 2025-08-18 at 11 33 47 PM

**TO DO:** I still need to write some brief user documentation on this. Should that be added to an existing page in the docs, and if so which one, or should it be a new page just for this? Suggestions welcome.

### DIFF
--- a/stpsf/mast_wss.py
+++ b/stpsf/mast_wss.py
@@ -896,6 +896,7 @@ def retrieve_iec_heater_telemetry(tstart, tend, plot=False):
 
 
     if plot:
+        import matplotlib.pyplot as plt
         fig, ax = plt.subplots(figsize=(16,4))
 
         for mnemonic, *other_vals in stpsf.constants.iec_mnemonics:
--- a/stpsf/mast_wss.py
+++ b/stpsf/mast_wss.py
@@ -862,3 +862,50 @@ def query_program_visit_times(program, verbose=False):
 
     # visit_times = np.asarray(visit_times)
     return astropy.table.Table([vids, starts, ends], names=('visit_id', 'start_mjd', 'end_mjd'))
+
+# Functions for retrieving telemetry about observatory wavefront variations, in particular the IEC thermal oscillations
+
+@functools.lru_cache
+def retrieve_iec_heater_telemetry(tstart, tend, plot=False):
+    """Retrieve JWST Instrument Electronics Compartment (IEC) heater telemetry, for use in de-trending WFSC
+
+    Parameters
+    ----------
+    tstart, tstop : strings
+        Datetimes for start and stop of time range
+
+    plot : bool
+
+    Make a plot, for test or debugging purposes.
+
+
+    """
+    import stpsf.constants
+
+    from jwst.lib.engdb_tools import ENGDB_Service
+    engdb_service = ENGDB_Service()  # By default, will use the public MAST service.
+
+    iecvals = {}
+
+    for mnemonic, amp, lag, t_mean in stpsf.constants.iec_mnemonics:
+        times, values = engdb_service.get_values(mnemonic, tstart, tend, include_obstime=True, zip_results=False)
+        if len(times)== 0:
+            raise RuntimeError(f"Found no results for {mnemonic} between {tstart} - {tend}")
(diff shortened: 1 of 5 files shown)


mperrin:
Ah, so the tests are failing because this PR uses a utility function from the `jwst` package (the pipeline) for retrieving the telemetry from MAST; I hadnt' realized we don't actually have the jwst pipeline as a dependency for STPSF.  So that usage is problematic slightly. 

I'm inclined to make this an optional dependency, and make the CI only test this function for the IEC WFE if that optional dependency is installed. We have until now not included the whole pipeline as a dependency for this and I think we should stick with that (not least because, this is a multi mission tool, and some Roman use cases will likely not want to install the JWST pipeline). At the same time I don't want to rewrite a version of the telemetry retrieval code that duplicates all of `ENGDB_Service` from the pipeline here either so I still want to use the version in jwst for that.

stpsf/mast_wss.py
```diff
@@ -862,3 +862,50 @@ def query_program_visit_times(program, verbose=False):
...
+    for mnemonic, amp, lag, t_mean in stpsf.constants.iec_mnemonics:
+        times, values = engdb_service.get_values(mnemonic, tstart, tend, include_obstime=True, zip_results=False)
+        if len(times)== 0:
+            raise RuntimeError(f"Found no results for {mnemonic} between {tstart} - {tend}")
+        times = astropy.time.Time([v.value for v in times])   # Convert from list of Times to one Time with an ndarray of values.
+        iecvals[mnemonic] = times, np.asarray(values)
+
+
+    if plot:
+        fig, ax = plt.subplots(figsize=(16,4))
```
marcio-melendez, line 900: When choosing` plot=True` this fails because it needs `import matplotlib.pyplot as plt` at the beginning of   mast_wss.py

marcio-melendez (commented):
I can reproduce the figure that you included on this PR with the example code. However, if I try to pass an array of times  it crashes when the time separation is too large. For example, anything a bit larger than this will crash in my computer
`stpsf.opds.estimate_iec_induced_wfe_at_time(['2024-08-18T00:00:00','2024-08-18T01:00:00','2024-08-18T04:00:00'], plot=True)`
On a notebook it terminates the kernel and on the terminal I get a segmentation fault: 11
It seems that anything larger than a ~4hr period will crash. 
You could add a quick check for the time range not to exceed ~4hrs. Note sure if people may want to use a large time range for a TSO or other large observation. 
image

mperrin:
Hmm, on my computer it runs fine for longer periods. Maybe some machine-dependent issue such as not enough RAM or something? Here's a plot for 24 hours: 
Screenshot 2025-08-21 at 12 13 53 PM

So I am inclined to not put in any artificial limit on maximum time interval. 

Thanks for the other feedback comments - I'll implement those fixes now.

marcio-melendez (approved):
Looks good. Thanks for the changes.

Issue #111, opened 2025-08-07
Roman charge diffusion & jitter


Skyhawk172:
I received communication from Goddard regarding recent measurements of the detector charge diffusion so we can go ahead and implement it for Roman, similar to how it's done for JWST.

In addition, the jitter expectations are lower than the currently-used values in STPSF, so we could also update the value.

> Based on the results in the attached paper, my recommendation would be to use a placeholder value of 3 microns (Gaussian sigma) [tldr: for charge diffusion]. This corresponds to ~33mas of pointing jitter.  Note that the STPSF docs state that the anticipated jitter is 12mas/axis; the reality is closer to 5-6 mas. The requirement of 8mas is exceeded in under 1% of the present simulations. My recommendation would be to reduce the default to at least the requirement value, or maybe lower.

mperrin:
Let me double check my understanding of what you wrote above. Am I correct that GSFC recommend to: 
1. For charge diffusion, model it as a Gaussian with sigma = 33 mas.
2. For pointing jitter, model it as a Gaussian with sigma = somewhere in the range 5 to 8 mas. 

If so, then that's qualitatively consistent with JWST, which likewise has the effect of charge transfer nearly an order of magnitude more than pointing jitter. (See, I told you it was important to get this into the models! ;-)  

Given the relative magnitudes, it makes very little difference whether you pick 5 or 6 or 8 mas as the default value for jitter; the charge diffusion is by far the dominant effect, and the net effect of the two convolutions ends up around 33.5 to 33.9 ish overall.

mperrin:
@ojustino For charge diffusion, the implementation is very straightforward: 
in `constants.py`, set 
```
INSTRUMENT_DETECTOR_CHARGE_DIFFUSION_DEFAULT_PARAMETERS['WFI'] = 0.033
```

marcio-melendez:
I add this small change to PR #118 . Once that PR is merge then I'll close this issue.

marcio-melendez:
PR #118 has been merged and includes, as a default,  the value for charge diffusion for Roman as discussed above. We'll open another issue for IPC effects once we have further details.

mperrin:
For the record:
@marcio-melendez's PR #118 implements the charge diffusion part of this. 
And I just created #119 for the LOS jitter Gaussian part

PR #118, opened 2025-09-19
added large psf pupil update for miri and minor charge difusion term …

marcio-melendez:
This PR enhance support for a large PSF calculation for MIRI by resampling the tricontagon amplitude, previously hardcoded to 1024x1024 in size. 
I am also using this PR to update Roman WFI charge diffusion value from zero to 0.033 mas (see #111)
Example code for the MIRI large PSF using a larger 4x JWST pupil:
```
miri = stpsf.NIRCam()
miri.pupil = 'jwst_pupil_RevW_npix4096.fits.gz'
miri.load_wss_opd_by_date('2024-08-01T00:00:00')  
single_stpsf_larger_pupil = miri.calc_psf(fov_pixels = 2048)   
```
image

### DIFF
--- a/stpsf/optics.py
+++ b/stpsf/optics.py
@@ -11,6 +11,7 @@ import poppy.utils
 from astropy.table import Table
 from scipy.interpolate import RegularGridInterpolator, griddata
 from scipy.ndimage import rotate
+from scipy.ndimage import zoom
 
 from . import constants, utils, stpsf_core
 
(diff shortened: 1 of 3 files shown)


marcio-melendez:
@ojustino and @Skyhawk172 FYI. I used this PR to add the charge diffusion value for Roman WFI as dicussed in issue #111 . This was a minor change so I included it on this PR.

stpsf/optics.py
```diff
@@ -1408,11 +1409,13 @@ def __init__(self, instrument, include_oversize=False, **kwargs):
...
                 self.amplitude = self.amplitude[256: 256 + 1024, 256: 256 + 1024]
             elif self.instrument.name == 'MIRI':
-                self.amplitude = fits.getdata(
+                amplitude_nominal = fits.getdata(
                     os.path.join(
                         utils.get_stpsf_data_path(), 'MIRI', 'optics', 'MIRI_tricontagon_oversized_rotated.fits.gz'
                     )
                 )
+                # this will adjust the amplitude size according to the opd size
+                self.amplitude = zoom(amplitude_nominal, npix / 1024)
```
mperrin: Consider making this:
```
self.amplitude = amplitude_nominal if npix==1024 else zoom(amplitude_nominal, npix / 1024)
```

That should avoid any slowdown from calling zoom unnecessarily  for the common typical case of 1024 pixels.

marcio-melendez: Yes, I investigated this before and zoom returns the same input array when zoom = 1 so there should not be any slowdown due to any calculation.  However, I noticed that our scipy version requirement range is outside this zoom implementation. I'll go ahead an adopt your change and update. 
Thanks!  
https://github.com/scipy/scipy/pull/22651/files

mperrin (approved):
Looks good to me.

PR #123, opened 2025-11-07
Misc improvements to trending plots, in particular for FP6 vs FP1

mperrin:
Several small improvements to `stpsf.trending.single_measurement_trending_plot` and `wfe_histogram_plot`

- The main one is better handling of the FP6 vs FP1 target OPDs, in particular for the post-MIMF OPD shown at the bottom row. We have to subtract the target phase map for FP1 from that one, even though we are now using FP6 for current measurements. The current (incorrect) usage of the FP6 target with the post-MIMF OPD at FP1 leads to systematic extra WFE in the display; this PR fixes that. 
- For corrections, make the code smarter about finding SUR files in the current directory. In particular if it can't find an exact match for the OPD from the sensing WAS AP file, it will look for other filenames matching in the date but differing in initial letter and/or final number. For example if it looks for `O2025103001_sur.xml` and doesn't find that, it will go on to correctly find `R2025103002_sur.xml`. (assuming the file has been downloaded from TAS)
- Also fixes the plot supertitle to correctly say "FP6" for sensing results using FP6. 
- For the histogram plot, if the time interval is > 3 years, reduce the X axis ticks to major ticks every 3 months and minor ticks every 1 month. This keeps from having a truly unreasonable number of ticks. 

Examples of `single_measurement_trending_plot`. Note the improvements in the first row 3rd panel, and in the bottom row panels 1, 2 and 4. 

**CURRENT CODE, BEFORE THIS PR**
Unknown-1

**IMPROVED, WITH THIS PR**
Unknown-3

### DIFF
--- a/stpsf/trending.py
+++ b/stpsf/trending.py
@@ -352,8 +352,13 @@ def wfe_histogram_plot(
 
     sensing_markers, = axes[0].plot_date(dates.plot_date, np.asarray(rmses) * 1e3, '.', ms=ms, ls='-',
                                          label='Sensing visit')
-    axes[0].xaxis.set_major_locator(matplotlib.dates.DayLocator(bymonthday=[1]))
-    axes[0].xaxis.set_minor_locator(matplotlib.dates.DayLocator(interval=1))
+    if end_date - start_date < 3*u.year:
+        axes[0].xaxis.set_major_locator(matplotlib.dates.DayLocator(bymonthday=[1]))
+        axes[0].xaxis.set_minor_locator(matplotlib.dates.DayLocator(interval=1))
+    else:
+        axes[0].xaxis.set_major_locator(matplotlib.dates.MonthLocator(bymonth=[1, 4, 7, 10]))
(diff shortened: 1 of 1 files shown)


stpsf/trending.py
```diff
@@ -704,6 +715,13 @@ def single_measurement_trending_plot(
             print('   That measurement has a correction, therefore reading in post-move sensing data and SUR.')
             print(f'   Inferred prior correction ID is {prior_corr_id}')
         sur_fn = f'{prior_corr_id}_sur.xml'
+        if not os.path.exists(sur_fn):
+            # check for slight differences in SUR fn
+            import glob
+            try_sur_fns = glob.glob('?'+sur_fn[1:7]+"*.xml")
```
marcio-melendez: It is better to add `*_sur.xml', otherwise the search will return a list with any xml file not only SUR, and the code will return an error when trying to use it. This happens in particular with the session data file which also have the same correction ID. Initially this happened to me because in the same folder I have the contents of the AP including the session data, the SUR and trail SUR, so the code was not picking the correct SUR.

marcio-melendez (approved):
Looks good!

PR #125, opened 2025-12-02
 Improve NIRCam coron SI WFE model, based on empirical testing

mperrin:
For NIRCam coronagraphy only, the SI WFE model is derived from optical modeling calculations in Zemax rather than ground test data. We've known this is imperfect for a while and there have been systematic issues in the PSF models that can easily be seen. 

Empirical testing and PSF comparisons by @kdlawson have recently shown that the PSF models can be significantly improved by simply flipping the SI WFE model in the vertical direction.  This PR implements such a flip, for NIRCam coron WFE _only_.   There are multiple ways such a fix could have been implemented; in this case I chose to do so with a change in code, rather than a change in the data file, because I judged that this is the easier way to roll out a change immediately and consistently to anyone using the `develop` version of STPSF. 

This only needs a pretty small & simple code change. 

Here's Kellen's discovery, copied here from a Slack discussion. 

@kdlawson wrote: 
> The first attached GIF flips between the final derotated image of HR 4796B in F444W+MASK335R and the best-fit nominal STPSF model. I'm not tuning distortion or pupil shear, etc — just source position and brightness. You can see the previously discussed halo/shadowing that differs significantly between the data and the nominal model.
The second GIF makes the same comparison but now using a model where I've literally just subclassed stpsf.NIRCam to flip the SI WFE OPD along the y-axis
> With no other tuning, chisq is reduced by a factor of ~2.5 with the flipped SI WFE model!

[Ed. note: Github is making these images huge, not sure why]

hr4796b_nominal_stpsf_fit_noresids
hr4796b_flippedsiwfe_stpsf_fit_noresids

The same improvement is also seen for NIRCam SW. Again from @kdlawson:
> SW definitely sees an improvement with the vertical SI WFE flip, at least. The "data" frame in each case is actually the empirical HPFRDI PSF model for HR 4796A, rather than the target itself (because the disk is very bright).
There's still some obvious inaccuracy in the flipped SI WFE version (2nd GIF), but it's worth noting that the latter is using the pupil shear fit with the nominal PSF models. Refitting pupil shear would likely clean this up even more. In any case, the shape of the speckles is obviously improved substantially here by flipping the SI WFE.

hr4796a_F200W_MASK335R_nominal_stpsf_fit_noresids
hr4796a_F200W_MASK335R_flippedsiwfe_stpsf_fit_noresids

### DIFF
--- a/stpsf/optics.py
+++ b/stpsf/optics.py
@@ -1434,6 +1434,15 @@ class WebbFieldDependentAberration(poppy.OpticalElement):
             self.opd = poppy.zernike.opd_from_zernikes(coeffs, npix=npix, outside=0)
             self.amplitude = (self.opd != 0).astype(int)
 
+        # Special case for NIRCam coronagraphy: we need to flip the OPD model in the Y axis.
+        #   In this case we are using a lookup table of Zernike coefficients derived from Zemax models
+        #   and it has been empirically determined that there seems to be a coordinate system inconsistency
+        #   in that. Flipping that vertically in the Y axis results in model PSFs that better match
+        #   observed coronagraphic PSFs.
+        if is_nrc_coron:
+            self.opd = self.opd[::-1]
+            self._is_OPD_flipped_for_NRC_coron = True
+
     def header_keywords(self):
         """Return info we would like to save in FITS header of output PSFs"""
         from collections import OrderedDict
--- a/stpsf/tests/test_nircam.py
+++ b/stpsf/tests/test_nircam.py
@@ -158,6 +158,11 @@ def do_test_nircam_blc(clobber=False, kind='circular', angle=0, save=False, disp
         else:
             raise ValueError("Don't know how to check fluxes for angle={0}".format(angle))
 
+    # we also verify that the SI WFE has been flipped in the Y axis in this case, as needed for NIRCam coronagraphy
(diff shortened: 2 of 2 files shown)


mperrin:
FYI @kdlawson @jarronL @AarynnCarter  @juliengirard

mperrin:
There is one test failure which still needs to be debugged on this - I posted on Slack to relevant folks some details. So that will need to be dealt with before this is ready to merge.

mperrin:
From @kdlawson via Slack: 
> Confirming that the PR effects match expectation!

i.e. Kellen has tested that this PR works consistently with his local tests, and improves the SI PSF model fidelity for NIRCam coronagraphy observations.

marcio-melendez (approved):
Small enough changes with important implications. I did the test and it is consistent with other comments in this PR; and all the tests are passing

juliengirard:
Awesome find! 
What do you think drives the biggest remaining differences? Pupil shear? Some dispersion?

mperrin:
@juliengirard 
> What do you think drives the biggest remaining differences? Pupil shear? Some dispersion?

Pupil shears, and imperfect models of the SI wavefront error overall at these field points. 
I think those are likely the dominant contributors.

PR #127, opened 2025-12-03
Improve load_wss_opd optional plot; add option to toggle which target phase map model is used

mperrin:
Inspired by slack discussion with @marcio-melendez on cross checking details of OTE and SI wavefront models. This PR replaces and expands the optional plot that can be shown in `load_wss_opd` and `load_wss_opd_by_date`, so now it more explicitly shows the different model components and directly states the math done to yield the resulting OPD. 

(This is in part code from an old branch I'd started in the webbpsf days, and had not yet moved over to STPSF. I started from that and did further updates and improvements now.)

A second addition is a new keyword argument to those functions, `use_exact_wss_target_phase`. This controls which model is used to subtract the SI wavefront from the sensing data, either (if True) the WSS target phase map, or (if False) the interpolated ISIM CV Zernikes, i.e. same model as used for SI WFE in general.  For now, True is the default, which is equivalent to the same behavior as used prior to this PR. But I think we might want to discuss whether this should be False by default. TBD. 

Example plots: 

```
nrc = stpsf.NIRCam()
nrc.aperturename = 'NRCA1_FP6'
nrc.load_wss_opd_by_date('2025-11-28', plot= True)
```
Unknown-2

And showing the effect of the new `use_exact_wss_target_phase` argument: 

```
nrc = stpsf.NIRCam()
nrc.aperturename = 'NRCA1_FP6'
nrc.load_wss_opd_by_date('2025-11-28', plot= True, use_exact_wss_target_phase=False)
```

Unknown-3

### DIFF
--- a/stpsf/stpsf_core.py
+++ b/stpsf/stpsf_core.py
@@ -1616,7 +1616,8 @@ class JWInstrument(SpaceTelescopeInstrument):
             self, slew_delta_time=slew_delta_time, slew_case=slew_case, ptt_only=ptt_only, verbose=verbose
         )
 
-    def load_wss_opd(self, filename, output_path=None, backout_si_wfe=True, verbose=True, plot=False, save_ote_wfe=False):
+    def load_wss_opd(self, filename, output_path=None, backout_si_wfe=True, verbose=True, plot=False, save_ote_wfe=False,
+                     use_exact_wss_target_phase=True):
         """Load an OPD produced by the JWST WSS into this instrument instance, specified by filename
 
         This includes:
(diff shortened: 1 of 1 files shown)


stpsf/stpsf_core.py
```diff
@@ -1655,6 +1656,13 @@ def load_wss_opd(self, filename, output_path=None, backout_si_wfe=True, verbose=
             Note that the exported OPD file will give the OTE estimated total WFE at the selected Instrument's field
             point, not the OTE global at master chief ray, since it is the OTE WFE at the selected field point
             which is most of use for some other tool.
+        use_exact_wss_target_phase : bool
+            WebbPSF contains two versions of the model for the NIRCam "target phase map" at field point 1 (FP1),
```
marcio-melendez: Replace WebbPSF with STPSF :) 
Otherwise it looks good
image

marcio-melendez (approved):
Looks good.

PR #128, opened 2025-12-10
update query range in MAST WSS OPD query for 4 day WFS cadence

mperrin:
Tiny PR. Updates the search range for finding OPDs to use a 2x larger search time period for recent dates, after the WFS cadence switched from 2 days to 4 days. 

The code works as is, but will often issue warnings like: 
```
WARNING: NoResultsWarning: Query returned no results. [astroquery.mast.discovery_portal]
iterating query, tdelta=3.0
```
This quiets that initial warning by just starting out with a ±3 days search range for recent dates.


marcio-melendez (approved):
Looks good

PR #132, opened 2026-01-05
[SCSB-225] migrate `[tool.pytest.ini_options]` to native TOML

zacharyburnett:
Resolves [SCSB-225](https://jira.stsci.edu/browse/SCSB-225)

Pytest 9.0 includes a native TOML table `[tool.pytest]`, which we should use instead of `[tool.pytest.ini_options]`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,7 @@ dependencies = [
 
 [project.optional-dependencies]
 test = [
-    "pytest",
+    "pytest>=9.0",
     "pytest-astropy",
 ]
 docs = [
@@ -70,8 +70,8 @@ namespaces = false
 local_scheme = "no-local-version"
 write_to = "stpsf/version.py"
 
-[tool.pytest.ini_options]
-minversion = "2.2"
+[tool.pytest]
+minversion = "9.0"
 testpaths = [
     "stpsf/tests",
 ]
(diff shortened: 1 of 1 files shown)


york-stsci:
This PR doesn't actually change any files, so as far as I can see it doesn't rename the pytest options (or add a pytest version pin) to pyproject.toml. Were those supposed to be included in the PR?

zacharyburnett:
> This PR doesn't actually change any files, so as far as I can see it doesn't rename the pytest options (or add a pytest version pin) to pyproject.toml. Were those supposed to be included in the PR?

oh, whoops! Fixed now

york-stsci:
One more question – earlier today we merged in PR 134, which makes changes to the dependencies in pyproject.toml. They shouldn't conflict with anything you've done, but I want to make sure you're not going to overwrite them. Otherwise it all looks good and is ready for approval.

zacharyburnett:
> One more question – earlier today we merged in PR 134, which makes changes to the dependencies in pyproject.toml. They shouldn't conflict with anything you've done, but I want to make sure you're not going to overwrite them. Otherwise it all looks good and is ready for approval.

sure thing, I just rebased this branch onto that change so there should be no conflicts

york-stsci (approved):
Looks good. Thank you!

PR #134, opened 2026-01-14
STPSF Dependency Updates

york-stsci:
- Synchronizes requirements.txt and pyproject.toml
- Adds upper version bounds to pyproject.toml


marcio-melendez (approved):
This looks good and in agreement with Slack conversations on the STPSF group. Thanks!

zacharyburnett:
sorry, I may just be ignorant of your workflow, but why do you have a `requirements.txt`? The requirements in `pyproject.toml` are sufficient to install requirements with `pip install`

mperrin:
I agree it's redundant to have both. Right now the requirements gets used by the GitHub CI so that needs to get reconfigured before deleting the requirements file.  I think we just need someone to update https://github.com/spacetelescope/stpsf/blob/develop/.github/workflows/ci_workflows.yml 
@york-stsci might that be something you could look at, sometime as time permits?

PR #135, opened 2026-01-14
remove `requirements.txt` and switch automated tests to reusable OpenAstronomy workflow

zacharyburnett:
> I agree it's redundant to have both. Right now the requirements gets used by the GitHub CI so that needs to get reconfigured before deleting the requirements file.  I think we just need someone to update https://github.com/spacetelescope/stpsf/blob/develop/.github/workflows/ci_workflows.yml 
@york-stsci might that be something you could look at, sometime as time permits?

_Originally posted by @mperrin in https://github.com/spacetelescope/stpsf/issues/134#issuecomment-3750988879_
            
I hope I'm not stepping on anyone's toes here, I was just already working in this repo so I figured I'd make a PR with this change


york-stsci:
It looks like this PR gets the configuration to work by removing the "latest" test entirely. Is that the case? If so, is it because it's not needed, or will we need to then add it back in with another method?

Also, do we do any sort of check to make sure that our minimum versions all still work?

zacharyburnett:
> It looks like this PR gets the configuration to work by removing the "latest" test entirely. Is that the case? If so, is it because it's not needed, or will we need to then add it back in with another method?
> 
> Also, do we do any sort of check to make sure that our minimum versions all still work?

the `-latest` toxenv was simply using the `requirements.txt` verbatim, which had upper pins (making it no different from a normal install). If we do want a truly `-latest` test then we can copy the current setup for `-legacy` in `tox.ini`, but unpin the versions

Right now, the `-legacy` toxenv defines static versions in `tox.ini`, with `poppy` set to the latest commit:
https://github.com/spacetelescope/stpsf/blob/df67fef9503c2e18adf85bcf9b5860acf3749937/tox.ini#L17-L22

mperrin:
The original intent was to test both the upper and lower bounds of what's supported. Now that `pyproject.toml` defines both a minimum and maximum version, is it possible to configure some test cases to use those for two different test cases, rather than have minimums hard-coded elsewhere for `-legacy`? I don't know how to do that in tox but maybe you do?

york-stsci:
The ideal for us would be if we could parse the toml file to create pins, one set with all of the minimum versions, and one with all of the maximum versions, especially so that we don't have to hardcode them and then keep them up to date.

york-stsci:
It looks like there's a tox module named tox-min-req that does exactly what we want. As far as I know, for all of our tests we just get (by default) the highest version available for everything (as in the most recent version below our pin), so I don't think we need a separate max test.

Would we be okay with adding a new tox module requirement?

york-stsci:
https://pypi.org/project/tox-min-req/

accompanied by

```
# In pyproject.toml
[tool.tox.ini_sections]
requires = ["tox-min-req"]
```

mperrin:
tox-min-req, nice.  It looks like someone has already invented the wheel that we need. Cool!

> Would we be okay with adding a new tox module requirement?

From my perspective, I think that's the kind of decision which y'all in ISEB should be the relevant authorities for, rather than me or Marcio for STPSF. If it's good with you then I think no objection from us. 👍

york-stsci:
From my perspective, this is an active pip project (it averages around one release per year, which is still reasonably active) and if it works for us there's no reason to find our own way of doing this. Tagging @BradleySappington for a second opinion, but I don't have an issue with using it.

BradleySappington:
@york-stsci - Going to first acknowledge that UV has a method for testing the minimum version inherently, but I don't think its worth migrating this repository to UV just for the sake of it as tox is working just fine.  That said, your proposal of having tox-min-req is a-ok with me!  Just make sure its an optional "test" dependency in pyproject.toml (this section already exists and is populated with two elements).

york-stsci:
Thanks for the mention of UV. If we think that's a better way forward in general, we might be able to do a multi-step transition. For example, the tox-uv module (https://pypi.org/project/tox-uv/#description) lets you run tox with uv as the base instead of virtualenv/pip, and it *does* seem that it would include UV's ability to specify highest/lowest compatible versions (and to make it explicit in our testing).

zacharyburnett:
> Thanks for the mention of UV. If we think that's a better way forward in general, we might be able to do a multi-step transition. For example, the tox-uv module (https://pypi.org/project/tox-uv/#description) lets you run tox with uv as the base instead of virtualenv/pip, and it _does_ seem that it would include UV's ability to specify highest/lowest compatible versions (and to make it explicit in our testing).

I can cheerlead for `tox-uv`, it makes environment creation much faster at basically no cost. We also have `minimum_dependencies` which we wrote to retrieve minimums without needing an extra tox module or using `uv`: https://github.com/OpenAstronomy/minimum_dependencies

here's how `romancal` does it in their `tox.ini` (calling it `-oldestdeps`):
https://github.com/spacetelescope/romancal/blob/06806e544cd3e45b6a652c366dbf155c59e385d4/tox.ini#L38-L49
```ini
deps =
    xdist: pytest-xdist
    ddtrace: ddtrace
    oldestdeps: minimum_dependencies
commands_pre =
    oldestdeps: minimum_dependencies romancal --filename requirements-min.txt
    oldestdeps: pip install -r requirements-min.txt
    stdevdeps: pip install -r requirements-dev-st.txt -U --upgrade-strategy eager
    devdeps: pip install -r requirements-dev-thirdparty.txt -U --upgrade-strategy eager
    pip list
commands =
    pytest \
```

york-stsci:
It seems as if tox-uv at least can do that without the need for anything else. From the pypi page,

> We use uv pip to install packages into the virtual environment. The behavior of this can be configured via the following options:
> 
> uv_resolution
> This flag, set on a tox environment level, informs uv of the desired [resolution strategy](https://github.com/astral-sh/uv/blob/0.1.20/README.md#resolution-strategy):
> highest - (default) selects the highest version of a package satisfying the constraints.
> lowest - install the lowest compatible versions for all dependencies, both direct and transitive.
> lowest-direct - opt for the lowest compatible versions for all direct dependencies, while using the latest compatible versions for all transitive dependencies.
> This is an uv specific feature that may be used as an alternative to frozen constraints for test environments if the intention is to validate the lower bounds of your dependencies during test executions.

york-stsci:
@zacharyburnett do you want to look into this, or would you like me to check out your branch and submit PRs to it for addition to this one?

zacharyburnett:
I apologize for the flurry of commits as I figure this out; I don't have the `stpsf` environment set up locally 😅

zacharyburnett:
This _should_ be ready now. I went ahead and changed the test CI to use the reusable OpenAstronomy workflow, which is kept updated upstream and centralizes maintenance for all our repositories. I set it up to use `tox-conda` to install conda deps, and `tox-uv` to use the `lowest-direct` resolution for the `-legacy` env.

Merging this PR will also require updating the branch protection rules in the repository settings to use the new names of the test jobs (using the toxenv names)

zacharyburnett:
Since this package is used downstream in the pipeline environments, I personally recommend keeping the number of pins in `pyproject.toml` to a minimum, which includes reducing the lower pins of dependencies to the lowest possible supported versions that still work, and unpinning upper pins on dependencies that do not **absolutely need** to be pinned

mperrin:
Looks like this PR from January got stalled but is ~ready to go. I think it slipped out of attention since no one is tagged specifically to do a review. @york-stsci would you be able to review this? (time permitting, non urgent, no rush -- we just probably shouldn't forget about it for another 3 months 😆 )

pyproject.toml
```diff
@@ -19,29 +19,29 @@ dynamic = [
 readme = "README.rst"
 requires-python = ">=3.10"
 dependencies = [
-    "astropy>=6.1.2,<7.1.0",
-    "astroquery>=0.4.7,<0.5.0",
+    "astropy>=5.1.0,<7.1.0",
```
york-stsci, line 22: Is the reduction of the required astropy version intentional, or the result of dependabot merges happening upstream?

zacharyburnett: I don't remember where this change came from, so I'll revert it

zacharyburnett:
is using `tox-conda` necessary for testing this package?

pyproject.toml
zacharyburnett: ah I see now; these changes are from the existing dependencies specified in the `-legacy` toxenv in `tox.ini`:
https://github.com/spacetelescope/stpsf/blob/13e7b5e009872330e3cb40a3a35343820fc735a5/tox.ini#L19-L22

Which "oldest supported dependencies" are correct, the ones in `tox.ini` or the ones in `pyproject.toml`?
https://github.com/spacetelescope/stpsf/blob/13e7b5e009872330e3cb40a3a35343820fc735a5/pyproject.toml#L22-L30

york-stsci: Via dependabot I've been updating pyproject.toml, so I'd go with those. I'm not sure how to get dependabot to see tox.ini, but I'll look into it.

york-stsci:
WRT tox conda, don't know.

pyproject.toml
zacharyburnett: why update the minimum pins via dependabot? Does `stpsf` now require features only in those versions?

mperrin: STPSF has little to no dependency on specific features of upstream packages at this point. It does not in general make use of much/any version specific recent changes in the upstream packages. It doesn’t and shouldn’t require using the absolutely latest version of any upstream package. The original intent back when setting this up was more like “STPSF should run on pretty much any currently-supported non-EOL’ed version of Python or astropy; it doesn’t care much”. But that general concept seems maybe not straightforward to turn into a config file.

zacharyburnett: ok that sounds good to me! I'll keep the pins at what they were in `tox.ini` then, since the tests against those lowest versions are passing right now

york-stsci: That was me forgetting what we were talking about. I don't think that we have dependabot doing that, unless we're in a situation where supporting a higher astropy version requires dropping support for (say) a lower numpy version. I do know that pyproject has been actively looked at recently, and I don't know that tox has been.

zacharyburnett: we also recently added a parameter in OpenAstronomy that adds toxenvs to test the package against all non-EOL'd versions of Python; I'll add that

zacharyburnett:
> WRT tox conda, don't know.

Alright, I'll try removing it and see if the tests still pass

zacharyburnett:
looks like removing `tox-conda` worked fine, so this PR is ready to go!

york-stsci (approved):
Given changes and discussion, looks good. Thank you!

mperrin:
@york-stsci @zacharyburnett This PR looks good to go, except some of the CI tests don't seem to have run. Is that intentional, or in any case understood enough to be something we can safely ignore and just merge this? Or should we kick the tests to try to get them re-run? Thanks

york-stsci:
It's not intentional. I have no idea why they haven't run, or how to re-run them. I can re-run the completed checks but those 5 seem to have somehow gotten lost.

zacharyburnett:
> @york-stsci @zacharyburnett This PR looks good to go, except some of the CI tests don't seem to have run. Is that intentional, or in any case understood enough to be something we can safely ignore and just merge this? Or should we kick the tests to try to get them re-run? Thanks

that's because the branch protection ruleset in this repo are expecting CI runs with specific names, but this PR changes those names (because it's using a reusable workflow it prepends `test / ` to the job names anyway). That's why they're marked as "Required" but they never start

Someone with admin access to this repo will have to force-merge this PR and then change the ruleset at https://github.com/spacetelescope/stpsf/settings/branches to remove the old job names and expect the new ones

york-stsci:
@zacharyburnett thank you! That makes a lot of sense. Given it, I'm going to merge this in.

york-stsci:
@zacharyburnett how do we need to update the branch protection rules in this case (or for the future)?

zacharyburnett:
> @zacharyburnett how do we need to update the branch protection rules in this case (or for the future)?

It should be at https://github.com/spacetelescope/stpsf/settings/branches (or, because GitHub is always changing things 🙄 https://github.com/spacetelescope/jwst/settings/rules). Then you go to the "Required Status Checks" entry and it'll have a list of CI jobs that you can remove (or add new ones) to be required before a PR can be merged normally

mperrin:
@york-stsci here's another one where it looks like you approved it as good to merge, but it hasn't been merged yet. Want me to go ahead and merge it, or is there a reason to wait? Thanks!

york-stsci:
Oops, this is the one where I was trying to look at the settings @zacharyburnett mentioned, but it didn't seem like we have any defined tests there at all, so I wasn't sure if we should merge this now or wait until I've figured out what's going on there. I'll give it a look today.

PR #136, opened 2026-01-20
Trending time series - add program IDs for cycle 4 maintenance programs

mperrin:
Trivial PR to add the list of cycle 4 WFS program IDs. The only effect of this is changing the color coding of the data points in `wavefront_time_series_plot` so the points show up blue (regular maintenance) instead of purple (MIMF and other non-routine WFS). 

It would probably make sense to revisit that color coding feature and just remove it entirely at this point, so as to not have to add a list of PIDs each cycle, but I'll leave that for another day.


marcio-melendez (approved):
Looks good. We need to find a way to get the relevant PIDs not hard coded.

PR #137, opened 2026-01-29
Update installation.rst

marcio-melendez:
small addition to instruction text for stsci users

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -84,6 +84,8 @@ You should now be able to successfully ``import stpsf`` in a Python session.
       export STPSF_PATH="/grp/stpsf/stpsf-data"
       export PYSYN_CDBS="/grp/hst/cdbs"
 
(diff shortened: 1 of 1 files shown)


mperrin (changes requested):
A belated review on this one. Since this is no longer recommended, would it be better to simply delete entirely this section of the documentation? Simplify where possible; less can be more. What do you think?

I'm OK with this change if you prefer. Either way OK with me.  

Very minor requested change:  note the sentence ends abruptly. Should be "users." with an s and a period at the end.

marcio-melendez:
Agree, to remove the entire "For STScI Users Only" section, right?

mperrin:
Yes that was my thought. If it's not recommended and more, and has caused confusion, then let's just take it out. 👍

marcio-melendez:
Done.

mperrin (approved):

mperrin:
Note for posterity: the CI failure on this is an unrelated upstream dependency issue. So this is OK to merge.

PR #140, opened 2026-04-18
Consider all available Zernike terms for WFI

ojustino:
In preparing for the next release, I found that an internal Roman function manually sets the number of Zernike terms considered in a simulation to 22. However, the Cycle 10 delivery (implemented in #100) increased the number of Zernike terms available per detector/filter/field point combination from 22 to 45.

I changed the Roman code such that it calculates the number of Zernike terms based on the number of such columns present instead of assuming a number.

### DIFF
--- a/stpsf/roman.py
+++ b/stpsf/roman.py
@@ -11,6 +11,7 @@ WARNING: This model has not yet been validated against other PSF
 import logging
 import os.path
 import pprint
+import re
 from pathlib import Path
 
 import astropy.units as u
@@ -35,7 +36,7 @@ class WavelengthDependenceInterpolator(object):
     points
     """
 
-    def __init__(self, n_wavelengths=16, n_zernikes=22):
+    def __init__(self, n_wavelengths=16, n_zernikes=45):
         self._n_wavelengths = n_wavelengths
         self._n_zernikes = n_zernikes
         self._aberration_terms = np.zeros((n_wavelengths, n_zernikes), dtype=np.float64)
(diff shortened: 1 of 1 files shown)


marcio-melendez:
@ojustino do you think it will be possible to set the number of Zernike coefficients from the Zernike files? i.e., count the number of columns with Zernike terms and use that as the number for the simulation. Otherwise, every time we change the file we'll need to manually change the terms.

stpsf/roman.py
```diff
@@ -238,15 +243,17 @@ def build_detector_from_table(number, zernike_table):
...
-        wavelength space"""
+    def build_wavelength_dependence(rows, n_zernikes):
+        """Build an interpolator object that interpolates `n_zernikes` Zernike
+        terms in wavelength space"""
         wavelengths = set(rows['wavelength'])
-        interpolator = WavelengthDependenceInterpolator(n_wavelengths=len(wavelengths), n_zernikes=22)
+        interpolator = WavelengthDependenceInterpolator(
+            n_wavelengths=len(wavelengths),
+            n_zernikes=n_zernikes
+        )
```
ojustino, line 253: HI @marcio-melendez, this change (plus the added line further up that defines `n_zernikes`) calculates the number of terms based on the number of "ZNN" columns (Z1, Z2, etc.) present in the Zernike CSV files.

marcio-melendez (approved):
Looks good, great.

mperrin:
@marcio-melendez approved this. The failing test is due to an unrelated upstream dependency issue, so let's go ahead and merge.

PR #141, opened 2026-04-20
use local workflow so we can publish wheels to PyPI without needing a token (using Trusted Publishers)

zacharyburnett:
since PyPI does not currently allow a Trusted Publisher to be a reusable workflow (i.e. the OpenAstronomy build workflow), instead we can just copy a hardened build + publish workflow here. That way, the configuration on PyPI's side can be set up so that this workflow (`build.yml`) is set up as a "trusted publisher" and then we never need to use or know the PyPI token

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,15 +1,50 @@
 name: build
 
 on:
-  release:
-    types: [ released ]
   pull_request:
+  release:
+    types: [released]
   workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


york-stsci (approved):
This looks good to me. We know why the one test is failing (incompatible python version matrices in testing development versions), and this isn't the place to try to resolve it, so approved from my end.

zacharyburnett:
after merging this, in order to publish to PyPI, you'll need to configure the PyPI project https://pypi.org/project/stpsf to add this repository (`https://github.com/spacetelescope/stpsf`) and build workflow (`build.yml`) to the list of Trusted Publishers

mperrin:
@york-stsci I see you approved this last week but it's not been merged yet. Is there any reason to hold off on the merge, or should we go ahead with it now?

york-stsci:
I've been trying to figure out how to adjust the required tests, but I don't know if we need to do it right now

york-stsci:
Sorry, my comment was for a different PR. I'll merge this now and then update our pypi settings.

PR #142, opened 2026-04-28
Add display_wfs_visit function to easily display SW and LW data from a WFS visit

mperrin:
Here's a function I wrote last fall but did not get around to PRing at that time. This makes a display plot of the image data from a given visit, including both the SW and LW data.  We usually are looking at the WFS data just zoomed in to the single target star of course, but occasionally it's useful to see a big picture view of everything in the field of view.   

I don't see this being used often, and I'm uncertain about whether to merge this into STPSF at all versus put elsewhere such as my misc_jwst repo. But perhaps this would be a worthwhile additional display page to include in the TAS, maybe? Not sure. Opinions welcome. 

Parameters allow selection of which WFS visit (if not specified, the most recent visit is shown by default), and whether to show the WLM8 images, WLP8, or both WLM8 and WLP8 in two plots. 

```
stpsf.trending.display_wfs_visit('V07464133001')
```
Unknown

### DIFF
--- a/stpsf/mast_wss.py
+++ b/stpsf/mast_wss.py
@@ -750,6 +750,12 @@ def get_visit_nrc_ta_image(visitid, verbose=True, kind='cal'):
 
     if verbose:
         print(f'TA filename: {filename}')
+    return get_mast_filename(filename)
+
+def get_mast_filename(filename):
+    """ Retrieve a specified file from MAST
+    Returns a FITS HDUList object in memory, without writing to disk
+    """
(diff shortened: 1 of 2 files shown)


marcio-melendez (commented):
This is very useful at least for us. Can you add `origin = 'lower'` so it looks more like the standard orientation? 
It also works well when trying our WFSCMB visits, it will take the L3 products.

marcio-melendez:
@kulpster85 This PR could be very useful for TAS, so we have a sequence that display all the products. We talked about this in the past but this is a convenient function.

marcio-melendez (approved):
Thanks for the changes, this looks great. 
Image

PR #145, opened 2026-05-18
move 'latest dev version' CI tests to Python 3.14

mperrin:
The 'latest development version' CI test case instances might as well run on the most recent Python release. 

(I'm also using this as a simple/trivial test case PR to test whether the whole CI system is working as we want it to right now; did the CI issues we were having with pysiaf upstream causing get resolved?)


mperrin:
This was superseded by other PRs addressing the issue, in particular #149

PR #149, opened 2026-06-08
update test workflow with job names and dynamic Python versions

zacharyburnett:
also closer aligns workflows with package template

with these static job names, we can add the following as "required status checks" in the repository ruleset, and then their Python versions will change with `requires-python`:
```
test / run tests (oldest supported Python, ubuntu-latest)
test / run tests with coverage (latest supported Python, ubuntu-latest)
```


york-stsci (approved):
Thank you!

PR #152, opened 2026-06-17
Set Roman pupil radius correctly, even if pupil files have zero padding

### LINKED ISSUE #151: Pupil radius is sometimes derived from padded array extent, not physical pupil radius (Roman WFI)
I ran into this issue while simulating defocused PSFs with Roman I-Sim in order to make flight-like data files we can run through our focus-diverse phase retrieval pipeline for commissioning. I'm getting significantly less defocused PSFs (about 14%) than expected compared to my own code.

I think the heart of the problem is that the `pupil_radius` property of `SpaceTelescopeInstrument` is derived from the *array size* of the pupil in the .fits file (like `pixelscale * array_size`) rather than the physical pupil size. The RST pupil files have some array padding around the clear aperture (a difference compared to the JWST pupil files!), so the pupil size is actually significantly smaller than what the array size would imply. The pupil diameter being calculated from the array size is about 2.552 m while the actual the nominal value is more like 2.36312 m (this is the value in the .fits file headers which I believe is derived from the GSFC Code V model and so should be pretty accurate to the as-built system). 

There is a hard coded class constant `RomanInstrument.PUPIL_RADIUS = 2.4 / 2.0` at `roman.py` line 288, so I suspect the problem may have already been recognized at some point but unfortunately not all code paths are using this radius value. 

Here's a walkthrough of the code path (for the record I'm looking at poppy v1.1.2 and spsf v2.2.0). The `FITSOpticalElement.pupil_diam` property in `poppy_core.py` (line 3215) computes the diameter as `pixelscale * array_size`:

```python
@property
def pupil_diam(self):
    """Diameter of the pupil (if this is a pupil plane optic)"""
    return self.pixelscale * (self.amplitude.shape[0] * u.pixel)
```

In `stpsf_core.py` line 450, `pupil_radius` is set from the first optical element's `pupil_diam` property:

```python
self.pupil_radius = pupil_optic.pupil_diam / 2.0
```

The incorrect radius is passed directly to `poppy.ThinLens` at `stpsf_core.py` lines 486–488:

```python
lens = poppy.ThinLens(
    name='Defocus', nwaves=defocus_waves,
    reference_wavelength=defocus_wavelength, radius=self.pupil_radius
)
```

Inside `ThinLens.get_opd()` (`poppy/optics.py` lines 2029–2037), the defocus Zernike is normalized so that ρ = 1 at `r = self.radius`:

```python
r_norm = r / self.radius.to(u.meter).value
defocus_zernike = ((2 * r_norm ** 2 - 1) *
                   (0.5 * self.nwaves * self.reference_wavelength.to(u.meter).value))
```

Because `self.radius` is the padded-array radius (1.276 m) rather than the physical pupil radius (2.36312 / 2 = 1.18156 m), the actual pupil edge only reaches

```
ρ_edge = r_phys / r_array = 1.18156 / 1.276 ≈ 0.926
```

The defocus term `(2ρ² − 1)` runs from −1 at the center to +1 at ρ = 1, so its peak-to-valley across an aperture that only reaches ρ_edge is

```
PV_realized = [(2·ρ_edge² − 1) − (−1)] · (0.5·N·λ) = ρ_edge² · (N·λ)
```

against a requested PV of `N·λ`. The realized defocus is therefore scaled by

```
PV_realized / PV_requested = ρ_edge² = (r_phys / r_array)² = (1.18156 / 1.276)² ≈ 0.857
```

i.e. about 14% short of the requested defocus. It would certainly affect anyone using `ThinLens`, and I could imagine there are other places where `SpaceTelescopeInstrument.pupil_radius` is being used instead of `RomanInstrument.PUPIL_RADIUS`, so there could be other problems that haven't been found yet.

My crude, temporary workaround has been to make this tweak to `stpsf_core.SpaceTelescopeInstrument.get_optical_system`: 

```python
# Before (generic code that uses the padded array extent):
self.pupil_radius = pupil_optic.pupil_diam / 2.0

# After (uses the Roman-specific class constant):
self.pupil_radius = self.PUPIL_RADIUS * units.meter
```

That's not a real long-term solution because it would break JWST simulations (I assume). It may be best to remove the assumption from `poppy` that the pupil fills the array and look to the header for the physical pupil file instead, but I don't know what other side effects that might have. Hopefully someone at STScI who's familiar with the whole family of code repositories can figure that out and decide what the best solution would be. If someone makes a development branch with a proposed fix I'll be happy to test it on my particular use case.

mperrin:
This should fix #151 reported by @bergkoet.

Currently, we are setting both a `RomanTelescope.PUPIL_RADIUS` class attribute (~correctly) and a `self.pupil_radius` instance attribute (incorrectly), which is both not working and also more complex than is necessary. This PR simplifies to just setting a `self.pupil_radius` value, correctly. Conveniently the `PUPIL_RADIUS` one was only used in one single place in the code so this is not a complex edit.

I moved the numeric value definition to `constants.py` which is a better place for it anyhow.

### DIFF
--- a/stpsf/constants.py
+++ b/stpsf/constants.py
@@ -373,6 +373,8 @@ JWST_SEGMENT_RADIUS = 1.517 / 2
 JWST_CIRCUMSCRIBED_DIAMETER = 6.603464  # meters. Outer corners of B segments
 JWST_INSCRIBED_DIAMETER = 5.47334  # meters. Middle corners of C segments
 
+ROMAN_PUPIL_DIAMETER = 2.36312  # meters
+
 JWST_TYPICAL_LOS_JITTER_PER_AXIS = 0.0008  # milliarcseconds jitter, 1 sigma per axis. = approx 1 mas rms radial, typically
 
 ROMAN_TYPICAL_LOS_JITTER_PER_AXIS = 0.0060  # milliarcseconds jitter, 1 sigma per axis. See https://github.com/spacetelescope/stpsf/issues/111
--- a/stpsf/roman.py
+++ b/stpsf/roman.py
@@ -225,7 +225,7 @@ def _load_wfi_detector_aberrations(filename):
         single_detector_info = zernike_table[zernike_table['sca'] == number]
         field_points = set(single_detector_info['field_point'])
         detector = FieldDependentAberration(
-            4096, 4096, radius=RomanInstrument.PUPIL_RADIUS,
+            4096, 4096, radius=constants.ROMAN_PUPIL_DIAMETER/2,
             name=f"Field Dependent Aberration (WFI{number:02d})"
         )
         for field_id in field_points:
@@ -285,7 +285,6 @@ class _RomanInstrumentOptionsDict(dict):
 
 @utils.combine_docstrings
 class RomanInstrument(stpsf_core.SpaceTelescopeInstrument):
-    PUPIL_RADIUS = 2.4 / 2.0
     """
     RomanInstrument contains data and functionality common to Roman
     instruments, such as setting the pupil shape
@@ -294,6 +293,9 @@ class RomanInstrument(stpsf_core.SpaceTelescopeInstrument):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+
+        self.pupil_radius = constants.ROMAN_PUPIL_DIAMETER / 2  # override default value set in super(), based on FITS file size
+                                                                # which may not be correct since Roman pupil files have some padding
 
         self.siaf = stpsf_core.get_siaf_with_caching('roman')
         self._aperturename = None
--- a/stpsf/roman.py
+++ b/stpsf/roman.py
@@ -294,8 +294,7 @@ class RomanInstrument(stpsf_core.SpaceTelescopeInstrument):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
-        self.pupil_radius = constants.ROMAN_PUPIL_DIAMETER / 2  # override default value set in super(), based on FITS file size
-                                                                # which may not be correct since Roman pupil files have some padding
+        self.pupil_radius = constants.ROMAN_PUPIL_DIAMETER / 2 * u.meter
 
         self.siaf = stpsf_core.get_siaf_with_caching('roman')
         self._aperturename = None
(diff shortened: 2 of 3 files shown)


bergkoet:
Thanks for the fast response!

This doesn't quite fix it because the `.calc_psf` method has the side effect of changing the pupil radius (by way of `.get_optical_system`). It happens here:
https://github.com/mperrin/stpsf/blob/dfad7477ffbb330177c7be792d1234063df4f6b4/stpsf/stpsf_core.py#L450

Here's a minimal example:
```python
import stpsf

WL = 1.577e-6         # F158 central wavelength [m]
DEFOCUS_WV = 0.25     # requested defocus [waves P-V]
EXPECTED_RADIUS_M = 2.36312 / 2  # physical entrance pupil radius [m]

stpsf.setup_logging()
wfi = stpsf.roman.WFI()
wfi.filter = 'F158'
wfi.options['defocus_waves'] = DEFOCUS_WV
wfi.options['defocus_wavelength'] = WL
found_radius_before_m = wfi.pupil_radius

psf, wavefronts = wfi.calc_psf(monochromatic=WL, return_intermediates=True)

found_radius_after_m = wfi.pupil_radius.to_value('m')
rho_max = EXPECTED_RADIUS_M/found_radius_after_m

print(f"\n")
print(f"Physical entrance pupil radius:   {EXPECTED_RADIUS_M:.4f} m")
print(f"wfi.pupil_radius before calc_psf: {found_radius_before_m:.4f} m")
print(f"wfi.pupil_radius after calc_psf:  {found_radius_after_m:.4f} m")
print(f"rho_max = physical / ThinLens radius:    {rho_max:.4f}")
print(f"Expected defocus underscale (rho_max^2): {rho_max**2:.4f}")
```

Result:
```
Physical entrance pupil radius:   1.1816 m
wfi.pupil_radius before calc_psf: 1.1816 m
wfi.pupil_radius after calc_psf:  1.2763 m
rho_max = physical / ThinLens radius:    0.9258
Expected defocus underscale (rho_max^2): 0.8571
```

mperrin:
@bergkoet OK let's try this again. Thanks for the minimal example code, and for pointing out the exact line causing the error; that was very helpful. With the latest commit just added to this branch, I have your test code passing:
```
rho_max = physical / ThinLens radius:    1.0000
Expected defocus underscale (rho_max^2): 1.0000
```

bergkoet:
> @bergkoet OK let's try this again. Thanks for the minimal example code, and for pointing out the exact line causing the error; that was very helpful. With the latest commit just added to this branch, I have your test code passing:
> 
> ```
> rho_max = physical / ThinLens radius:    1.0000
> Expected defocus underscale (rho_max^2): 1.0000
> ```

That works for me, too. Many thanks!

ojustino (approved):

Issue #153, opened 2026-06-22
stpsf.trending.monthly_trending_plot fails due to matplotlib update


sbetti22:
I ran stpsf.trending.monthly_trending_plot(2024, 6, verbose=False) which caused this error:
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[5], line 1
----> 1 stpsf.trending.monthly_trending_plot(2024, 6, verbose=False)

File ~/miniconda3/lib/python3.13/site-packages/stpsf/trending.py:1399, in monthly_trending_plot(year, month, verbose, instrument, filter, vmax, pid, opdtable, shift_dates_offset)
   1395 fig.suptitle(f'WF Trending for {year}-{month:02d}{title_extra}', fontsize=fs * 1.5, fontweight='bold')
   1397 # Plot 1: Wavefront Error
-> 1399 axes[0].plot_date(dates_array.plot_date, rms_obs * 1e9, color='C1', ls='-', label='Observatory WFE at NIRCam NRCA3')
   1400 axes[0].plot_date(dates_array.plot_date, rms_ote * 1e9, color='C0', ls='-', label='Telescope WFE')
   1401 for ax in axes:

AttributeError: 'Axes' object has no attribute 'plot_date'
```

I used stpsf 2.2 and Matplotlib 3.11.  

In Matplotlib 3.11, plot_date has been completely removed https://matplotlib.org/stable/api/prev_api_changes/api_changes_3.11.0.html#plot-date 
According to the documentation, plot now works with datetime and numpy objects and can be replaced.  Hopefully a simple fix to update!

mperrin:
@sbetti22 thanks. As a potential simple fix please try:
```
axes[0].xaxis.axis_date()
axes[0].plot(dates_array.plot_date, rms_obs * 1e9, color='C1', ls='-', label='Observatory WFE at NIRCam NRCA3')
axes[0].plot(dates_array.plot_date, rms_ote * 1e9, color='C0', ls='-', label='Telescope WFE')
```

sbetti22:
That seemed to work!  I have opened PR [154](https://github.com/spacetelescope/stpsf/pull/154) to implement it.

mperrin:
Resolved by #154

PR #154, opened 2026-06-30
changing plot_date to plot due to matplotlib3.11 update

sbetti22:
This is a PR to fix the bugs raised in #153.  `plot_date` has been removed in Matplotlib 3.11.  According to the documentation, `plot` now works with datetime and numpy objects and can be replaced. 

Wherever `plot_date` was given in trending.py, I have replaced it with 

```
ax.xaxis.axis_date()
ax.plot()
```

This fix has been tested on both Matplotlib 3.11 and an earlier version of Matplotlib (3.10.9) and I confirmed that using plot produces the same results and is identical to the previous plot_date method in both versions of Matplotlib.

### DIFF
--- a/stpsf/trending.py
+++ b/stpsf/trending.py
@@ -132,41 +132,39 @@ def wavefront_time_series_plot(
     is_routine = np.asarray([int(v[1:6]) in routine_pids for v in opdtable[where_pre]['visitId']])
 
     # Plot all, with connecting line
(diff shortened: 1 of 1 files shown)


marcio-melendez (approved):
This works as intended without errors for matplotlib  3.11. Thanks for implementing this fix! Looks good. 

Image

PR #156, opened 2026-07-21
Add display of expected ("floor") RMS WFEs to wfe_histogram_plot

mperrin:
This adds display of the WAS-computed "expected" RMS WFE, i.e. the expected WFE after a hypothetical correction, to the display in `wfe_histogram_plot`. Similar to the existing plot of expected WFE included in the TAS RMSWFE plot. 

See gray line added in the top panel:

Unknown-32

I chose to make it a gray line so it's present but relatively visually unobtrusive; this is worthwhile data to include for context but I want to keep the visual emphasis on the measured RMS WFE values.


marcio-melendez (approved):
This looks good, it is a nice addition. 

Image

PR #159, opened 2026-07-27
exclude tests using remote data (e.g. MAST) from CI builds

mperrin:


york-stsci (approved):
This looks like a simple way to handle the issue.

PR #161, opened 2026-08-04
Fix broken API links in Roman notebook

ojustino:
Besides these, I didn't find any other occurrences of the outdated "pythonhosted" URLs in the `docs` directory.


mperrin (approved):
looks good to me. Good catch.  Wow, this is some old text with outdated URLs :-)

PR #144, opened 2026-05-13
Update Roman to use latest preflight (post-SCIPA) pupil maps. 

mperrin:
-  [x] Update pupil filenames
-  [ ] Make pupil auto update on detector position changes, too
- [ ] Select correct slice of the datacubes
- [x] Remove pupil controller.
- [ ] Test

PR #162, opened 2026-08-05
Adapt Roman code to new required data files

ojustino:
Cleans up my test branch with code changes to handle the new structure of the required Roman data from GSFC – OPDs, pupils, and expanded Zernike coefficients.

Let's see if changing `DATA_VERSION_MIN` in `stpsf/__init__.py` is enough to enforce the use of the new files as we test this branch. If we merge this branch before the release is ready, we may need to revert this change so others who want to use the `develop` branch but don't have the files can still run STPSF.

I included @mperrin's quick fix to prevent large slowdowns from the number of Zernike coefficients exceeding the `lru_cache`'s default maximum size, but we could still benefit from either aligning the calculation more with the approach taken in STPSF's JWST models (future task) or by making the `R()` function from poppy's `zernike.py` more efficient.

### DIFF
--- a/stpsf/stpsf_core.py
+++ b/stpsf/stpsf_core.py
@@ -448,8 +448,10 @@ class SpaceTelescopeInstrument(poppy.instrument.Instrument):
             self._extra_keywords.update(pupil_optic.header_keywords())
 
         # add coord transform from entrance pupil to exit pupil
-        # unless ref data are already in exit pupil orientation (like with WFI)
-        if self.name != 'WFI':
+        # (WFI: invert both axes until pupil orientations fixed in source files)
+        if self.name == 'WFI':
+            optsys.add_inversion(axis='both', name='OTE exit pupil', hide=True)
+        else:
             optsys.add_inversion(axis='y', name='OTE exit pupil', hide=True)
 
         # add rotation at this point, if present - needs to be after the
--- a/stpsf/stpsf_core.py
+++ b/stpsf/stpsf_core.py
@@ -448,7 +448,9 @@ class SpaceTelescopeInstrument(poppy.instrument.Instrument):
             self._extra_keywords.update(pupil_optic.header_keywords())
 
         # add coord transform from entrance pupil to exit pupil
-        optsys.add_inversion(axis='y', name='OTE exit pupil', hide=True)
+        # unless ref data are already in exit pupil orientation (like with WFI)
+        if self.name != 'WFI':
+            optsys.add_inversion(axis='y', name='OTE exit pupil', hide=True)
 
         # add rotation at this point, if present - needs to be after the
         # exit pupil inversion.
@@ -589,10 +591,15 @@ class SpaceTelescopeInstrument(poppy.instrument.Instrument):
                 pupil_transmission = self.pupil
             else:
                 raise TypeError('Not sure what to do with a pupil of ' 'that type: {}'.format(type(self.pupil)))
+            # ---- check if we have an index into a datacube; this is used for Roman
+            transmission_index = (self._pupil_datacube_index
+                                  if hasattr(self, '_pupil_datacube_index')
+                                  else None)
             # ---- apply pupil intensity and OPD to the optical model
             pupil_optic = poppy.FITSOpticalElement(
                 name='{} Entrance Pupil'.format(self.telescope),
                 transmission=pupil_transmission,
+                transmission_index=transmission_index,
                 opd=opd_map,
                 planetype=poppy.poppy_core.PlaneType.pupil,
                 # rotation=self._rotation
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -31,6 +31,10 @@ jobs:
       - retrieve_cache
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/tox.yml@4193751d511425d4edc1d5657c24b2128d49b017 # v2.6.4
     with:
+      setenv: |
+        STPSF_PATH: ${{ needs.retrieve_cache.outputs.cache_path }}/stpsf-data/
+      cache-path: ${{ needs.retrieve_cache.outputs.cache_path }}
+      cache-key: ${{ needs.retrieve_cache.outputs.cache_key }}
       envs: |
         - name: Check for Sphinx doc build errors
           linux: py312-docbuild
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -105,8 +105,6 @@ jobs:
         - name: run tests with poppy dev (latest support Python, ubuntu-latest)
           linux: py3-poppydev-xdist
           python-version: ${{ needs.supported-pythons.outputs.latest }}
-      fill: true
-      fill_platforms: linux
-      fill_factors: xdist
+        - linux: py3*-xdist
       artifact-path: |
         results.xml
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -22,16 +22,18 @@ env:
   PYTHON_VERSIONS: 3.9 3.10 3.11 3.12 3.13 3.14
 
 jobs:
+  retrieve_cache:
+    uses: spacetelescope/stpsf/.github/workflows/retrieve_cache.yml@ee1ced313cc7277f07226f03e64d038f3929d7a3 # setup retrieve_cache with development or latest_release
+    with:
+      dataset_type: development   # or latest_release
   check:
+    needs:
+      - retrieve_cache
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/tox.yml@4193751d511425d4edc1d5657c24b2128d49b017 # v2.6.4
     with:
       envs: |
         - name: Check for Sphinx doc build errors
           linux: py312-docbuild
-  retrieve_cache:
-    uses: spacetelescope/stpsf/.github/workflows/retrieve_cache.yml@ee1ced313cc7277f07226f03e64d038f3929d7a3 # setup retrieve_cache with development or latest_release
-    with:
-      dataset_type: development   # or latest_release
   supported-pythons:
     runs-on: ubuntu-latest
     steps:
@@ -65,7 +67,7 @@ jobs:
     outputs:
       python-version-json: ${{ steps.versions.outputs.json }}
       oldest: ${{ steps.versions.outputs.oldest }}
-      latest: ${{ steps.versions.outputs.latest }} 
+      latest: ${{ steps.versions.outputs.latest }}
   test:
     needs:
       - retrieve_cache
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -29,7 +29,7 @@ jobs:
   check:
     needs:
       - retrieve_cache
-    uses: OpenAstronomy/github-actions-workflows/.github/workflows/tox.yml@4193751d511425d4edc1d5657c24b2128d49b017 # v2.6.4
+    uses: OpenAstronomy/github-actions-workflows/.github/workflows/tox.yml@22ddf09dfabb443adddfb9f861b7d41a787d6b1a # v3.0.3
     with:
       setenv: |
         STPSF_PATH: ${{ needs.retrieve_cache.outputs.cache_path }}/stpsf-data/
@@ -76,7 +76,7 @@ jobs:
     needs:
       - retrieve_cache
       - supported-pythons
-    uses: OpenAstronomy/github-actions-workflows/.github/workflows/tox.yml@4193751d511425d4edc1d5657c24b2128d49b017 # v2.6.4
+    uses: OpenAstronomy/github-actions-workflows/.github/workflows/tox.yml@22ddf09dfabb443adddfb9f861b7d41a787d6b1a # v3.0.3
     secrets:
       CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
     with:
@@ -92,7 +92,7 @@ jobs:
           linux: py3-cov-xdist
           python-version: ${{ needs.supported-pythons.outputs.latest }}
           pytest-results-summary: true
-          coverage: codecov
+          coverage: github
         - name: run tests (latest supported Python, ubuntu-latest)
           linux: py3-xdist
           python-version: ${{ needs.supported-pythons.outputs.latest }}
--- a/stpsf/__init__.py
+++ b/stpsf/__init__.py
@@ -40,7 +40,7 @@ if sys.version_info < tuple(
 # required. If changes to the code and data mean STPSF won't work
 # properly with an old data package, increment this version number.
 # (It's checked against $STPSF_DATA/version.txt)
-DATA_VERSION_MIN = (2, 1, 0)
+DATA_VERSION_MIN = (2, 3, 0)
 
 
 class Conf(_config.ConfigNamespace):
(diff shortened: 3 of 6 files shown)


mperrin:
Hi @ojustino, I'm finding that many things in `test_roman.py` are failing with these changes. Are you able to successfully run the unit tests yourself?  Unless perhaps I've misconfigured something, it seems like we need some combination of test updates, some additional reference files, and/or a bit of debugging. 

Since the provided `prelaunch-ref-data` folder only contains WFI info, I set things up starting with a regular copy of the existing STPSF data, deleted the old WFI folder, and added in the new WFI folder and other files. Let me know if I should take a different approach? 

Here's the errors I get:
```

> pytest test_roman.py
[...]
test_roman.py .F.FF.FFFF...                                                                                                                               [100%]
[...]
[many lines of error traces omitted]
[...]


==================================================================== short test summary info ====================================================================
FAILED test_roman.py::test_WFI_filters - FileNotFoundError: [Errno 2] No such file or directory: '/Users/mperrin/software/webbpsf-data/WFI/aberrations/PRISM_5fields_Z45_multiwave.csv'
FAILED test_roman.py::test_WFI_fwhm - ValueError: pupilopd is set automatically on updates to the detector attribute
FAILED test_roman.py::test_WFI_pupil_controller - FileNotFoundError: [Errno 2] No such file or directory: '/Users/mperrin/software/webbpsf-data/WFI/aberrations/PRISM_5fields_Z45_multiwave.csv'
FAILED test_roman.py::test_WFI_includes_aberrations - AssertionError: Third plane of Roman WFI optical system should be the field dependent aberration virtual optic
FAILED test_roman.py::test_swapping_modes - FileNotFoundError: [Errno 2] No such file or directory: '/Users/mperrin/software/webbpsf-data/WFI/aberrations/PRISM_5fields_Z45_multiwave.csv'
FAILED test_roman.py::test_custom_aberrations - FileNotFoundError: [Errno 2] No such file or directory: '/Users/mperrin/software/webbpsf-data/WFI/aberrations/GRISM0_5fields_Z45_multiwave.csv'
FAILED test_roman.py::test_WFI_limits_interpolation_range - KeyError: 'wavelength'
================================================================= 7 failed, 6 passed in 36.25s ==================================================================

```

marcio-melendez:
It seems that the aberration files for prism and grism are missing. We did not do anything with these modes, i.e., created a multi-field 5x5 datafile. So, @ojustino , I think it is save for these two modes only to get the information from the 5 field points Goddard spreadsheet with Z45. In other words, update the files with the latest data from Goddard without any code change.

ojustino:
Hi @mperrin, sorry for missing your message earlier this week. I see the same test failures but am not expecting the branch to pass at this point. @marcio-melendez wanted this branch up as a PR for better visibility so others could begin simulating PSFs with the new required data. My plan was to update the tests addressing any feedback.

I should have made the PR a draft and will do that now.

mperrin:
Tasks - some brief notes from discussions between me, @marcio-melendez, and @kdlawson

- [x] Create new zip files, get them on Box. 
- [x] Update the CI to use those. Ensure the CI is passing here. 
- [x] Let's make sure the field point mapping code is tested and verified. In my previous draft PR #144, from which some code was copied, I had a function `test_sci_xy_to_fp` (see [here](https://github.com/spacetelescope/stpsf/pull/144/changes#diff-f10341680eb43103eca747c72beb523e9bb74bb16d0f6320232933387bf2d23aR322)). That test function should be copied into this branch. 
- [ ] Update docs, in particular regarding coordinate frames (this may be a separate PR?)

stpsf/stpsf_core.py
```diff
@@ -448,6 +448,7 @@ def get_optical_system(self, fft_oversample=2, detector_oversample=None, fov_arc
             self._extra_keywords.update(pupil_optic.header_keywords())
 
         # add coord transform from entrance pupil to exit pupil
+        # (WFI: inversion needed here until GSFC correctspupil orientations)
```
marcio-melendez: The sign-flip should be both X- and Y-. Currently there is only the Y-axis flip. For example:

Image

but we want the OPD at the exit pupil orientation to look like this, e.g., x- and y:

Image

stpsf/stpsf_core.py
```diff
@@ -448,7 +448,10 @@ def get_optical_system(self, fft_oversample=2, detector_oversample=None, fov_arc
             self._extra_keywords.update(pupil_optic.header_keywords())
 
         # add coord transform from entrance pupil to exit pupil
+        # (WFI: invert both axes until pupil orientations fixed in source files)
         optsys.add_inversion(axis='y', name='OTE exit pupil', hide=True)
+        if self.name == 'WFI':
+            optsys.add_inversion(axis='x', name='OTE exit pupil', hide=True)
```
marcio-melendez: I see the the coordinate plane change applied per axis, which results on having the two extra planes, see image. I think we can add the label 'both' to the plane, i.e., `optsys.add_inversion(axis='both', name='OTE exit pupil', hide=True) `

Image

ojustino: Got it. I've substituted in your suggested edit.

marcio-melendez: Ok. This looks good now
image

marcio-melendez:
> Tasks - some brief notes from discussions between me, @marcio-melendez, and @kdlawson
> 
> * [x]  Create new zip files, get them on Box.
> * [x]  Update the CI to use those. Ensure the CI is passing here.
> * [x]  Let's make sure the field point mapping code is tested and verified. In my previous draft PR [Update Roman to use latest preflight (post-SCIPA) pupil maps.  #144](https://github.com/spacetelescope/stpsf/pull/144), from which some code was copied, I had a function `test_sci_xy_to_fp` (see [here](https://github.com/spacetelescope/stpsf/pull/144/changes#diff-f10341680eb43103eca747c72beb523e9bb74bb16d0f6320232933387bf2d23aR322)). That test function should be copied into this branch.
> * [ ]  Update docs, in particular regarding coordinate frames (this may be a separate PR?)

@ojustino did you check Marshall action about the mapping code?

ojustino:
All of the tests work for me locally now after the latest commits, which include the science pixel to field point test. (Note that tests will still fail on this PR until we start the pre-release procedure for the required data on Box.)

I rebased this branch to bring in recent merges to `develop`, so watch out for that when you pull the latest changes. I also had to update to the `GRISM1` Zernike file, so re-download the test data zip from Box as well.

kdlawson:
Per suggestion of @marcio-melendez: for the sake of documentation, I'm summarizing some tests I reported on Slack below. **But, in short, the PR's implementation for generating OPDs for WFI looks sound.**

### Coordinates:

When using the pre-release Roman/WFI SIAF XML file, converting the provided OPD sample coordinates to the `SCI` frame gives a close match to the sample positions as encoded in the most recent version of the updated STPSF reference data (`WFI/aberrations/*_25fields_zernike_Z210_multiwavelength.csv` files).

image

### OPD grid interpolation:

When generating interpolated OPDs in STPSF for the field points where samples were provided by GSFC, results are in close agreement with provided samples (average RMS of 12±2 nm across the 450 field points plotted below). Results are comparable in other filters.

image


Finally, running the same test with a fully distinct PCA-based interpolation approach (that would be impractical to use in practice) gave similar performance to the adopted Zernike model approach. As such, I think that the residuals above are simply the cost of distilling the large input dataset to something shippable. While shipping the full dataset could produce perfect matches here (though still imperfect elsewhere), carrying out 2D interpolations of a sequence of 450 1024x1024 OPD maps is prohibitively expensive for the vast majority of users.

Great work on this PR!

marcio-melendez:
When running the file locally, with the latest stpsf-data I am getting some errors in pytest because the folder FGS is missing from the zip file. @ojustino how are you creating the release data, using the scripts inside dev_utils? 
Also, I am getting a couple of errors. Also, the zip file has the folder MAST_JWST_WSS_OPDs, which should not exist in the stpsf-data.zip release. 
```
FAILED tests/test_roman.py::test_WFI_filters - TypeError: unhashable type: 'MaskedConstant'
FAILED tests/test_roman.py::test_WFI_pupil_controller - TypeError: unhashable type: 'MaskedConstant'
FAILED tests/test_roman.py::test_custom_aberrations - TypeError: unhashable type: 'MaskedConstant'
```
do you also see those when running pytest locally?

ojustino:
@kdlawson Thank you for the visualization of the changes to the SIAF and your analysis of the new residuals. 

@marcio-melendez Did you re-download the data files from my Box link? That should resolve the WFI-specific errors you listed. Since this PR should only affect Roman-related code, I don't think the other failures will persist once we begin the pre-release process I mentioned. If they do, I think they'll be better handled in a PR with tests that use the updated files.

marcio-melendez:
With the updated files all Roman tests are passing, locally. I still have other test failing because it is not the "officila" data release, but since those are under JWST those should pass without issues.

kdlawson:
Ahah! I think I figured out the weird corner residuals for some SCAs in the 2nd figure I posted above (upper right samples, as displayed). 

In `get_aberration_terms` for `stpsf.roman.FieldDependentAberration`, if the point falls out of bounds of the samples (which happens sometimes here because, even though we're trying to use the exact positions, "detector_position" is always interpreted as int), the first fallback is to interpolate the two nearest points. I'm guessing this code was written for an older five point grid, so it attempts to truncate the field points to just the corners by doing:

```corners = field_points[1:]```

But for the 5x5 grids, this actually drops one of the corner samples. So, we end up grabbing the two "nearest" points from the selection labeled 'corners' below, and then interpolating those to approximate the OPD for the set position.

image

I think the fix here is to just not attempt to use only the corner points and instead compute distances to all field points and grab the nearest two from those. Here's a quick revised version of this method:

```
    def get_aberration_terms(self, wavelength):
        """Supply the Zernike coefficients for the aberration based on
        the wavelength and pixel position on the detector"""
        if self.field_position in self._wavelength_interpolators:
            # short path: this is a known point
            interpolator = self._wavelength_interpolators[self.field_position]
            coefficients = interpolator.get_aberration_terms(wavelength)
        else:
            # get aberrations at all field points
            field_points, aberration_terms = [], []
            for field_point_coords, point_interpolator in self._wavelength_interpolators.items():
                field_points.append(field_point_coords)
                aberration_terms.append(point_interpolator.get_aberration_terms(wavelength))
            aberration_array = np.asarray(aberration_terms)
            assert len(aberration_array.shape) == 2, (
                'computed aberration array is not 2D ' '(inconsistent number of Zernike terms ' 'at each point?)'
            )
            field_position = tuple(self.field_position)
            coefficients = griddata(np.asarray(field_points), np.asarray(aberration_terms), field_position, method='linear')
            if np.any(np.isnan(coefficients)):
                # FIND TWO CLOSEST INPUT GRID POINTS:
                dist = np.hypot(*(np.asarray(field_points) - np.asarray(field_position)).T)
                min_dist_indx = np.argsort(dist)[:2]  # keep two closest points
                # DEFINE LINE B/W TWO POINTS, FIND ORTHOGONAL LINE AT POINT OF INTEREST,
                # AND FIND INTERSECTION OF THESE TWO LINES.
                x1, y1 = field_points[min_dist_indx[0]]
                x2, y2 = field_points[min_dist_indx[1]]
                dx = x2 - x1
                dy = y2 - y1
                a = (dy * (field_position[1] - y1) + dx * (field_position[0] - x1)) / (dx * dx + dy * dy)
                closest_interp_point = (x1 + a * dx, y1 + a * dy)
                # INTERPOLATE ABERRATIONS TO CLOSEST INTERPOLATED POINT:
                coefficients = griddata(
                    np.asarray(field_points), np.asarray(aberration_terms), closest_interp_point, method='linear'
                )
                # IF CLOSEST INTERPOLATED POINT IS STILL OUTSIDE THE INPUT GRID,
                # THEN USE NEAREST GRID POINT INSTEAD:
                if np.any(np.isnan(coefficients)):
                    coefficients = aberration_terms[min_dist_indx[0]]
                    _log.warn(
                        'Attempted to get aberrations at field point {} which is outside the range '
                        'of the reference data; approximating to nearest input grid point'.format(field_position)
                    )
                else:
                    _log.warn(
                        'Attempted to get aberrations at field point {} which is outside the range '
                        'of the reference data; approximating to nearest interpolated point {}'.format(
                            field_position, closest_interp_point
                        )
                    )
                assert not np.any(np.isnan(coefficients)), 'Could not compute aberration ' 'at field point {}'.format(
                    field_position
                )
        if self._omit_piston_tip_tilt:
            _log.debug('Omitting piston/tip/tilt')
            coefficients[:3] = 0.0  # omit piston, tip, and tilt Zernikes
        return coefficients
``` 

Here's the test figure using this version:

image

marcio-melendez:
that's great! @kdlawson can you make commits to this PR? @ojustino you may need to give permission to Kellen. Otherwise, if this is the function that needs to be replaced then Justin may be able to do so, and I can do the test similar to what Kellen did above.

kdlawson:
I cannot commit. If it's easier than giving access, you can just copy-paste the above in instead.

ojustino:
I can do that. You currently have "triage" permissions in the repository and are listed as an outside collaborator, so that should be updated. @marcio-melendez is there a formal process for that now?

ojustino:
This is a cleaner approach, even for the remaining 5-point filters (grisms and prism). filters. Thanks, @kdlawson.

marcio-melendez:
> I can do that. You currently have "triage" permissions in the repository and are listed as an outside collaborator, so that should be updated. @marcio-melendez is there a formal process for that now?

@kdlawson I think you can open an IT ticket requesting the necessary permissions for both STPSF and Poppy repositories.

mperrin:
(Popping up briefly, still mostly out this week)
I changed the permissions for Kellen. You could too Marcio, just go to the Settings page in the repository view on GitHub. 

It does still list Kellen as an “outside collaborator” - I think that’s because Kellen you may have not yet had your account added into the spacetelescope organization?

kdlawson:
Thanks, Marshall! That makes sense. I've just had my account added, so that will hopefully fix things.

marcio-melendez:
@ojustino is this ready for approval? I see 3 skipped tests but everything else seems fine. There is also [PR#166](https://github.com/spacetelescope/stpsf/pull/166), which has been reveiwed already. Is the idea to merge this PR to dev and then rebased PR166 (IPC) and then to merge?

ojustino:
The latest commits 1) ensure that the docbuild test uses the prerelease Box data file instead of the latest, and 2) prevent a code coverage test from being skipped.

The former makes it such that when `develop` PRs require updated paths in the required data, tests can actually pass by accounting for them appropriately. The latter appears as a Markdown table on the summary page of each GitHub Actions run ([example here](https://github.com/spacetelescope/stpsf/actions/runs/34638663568?pr=162#summary-103395765524)).

ojustino:
@marcio-melendez Yes, that's the idea. If you're OK with the non-GitHub Actions changes, you can give a formal approval when you're ready.

marcio-melendez (approved):
All the functional changes looks good to me. This is the culmination of a large set of tests, reviews and changes, some of which are attached to this PR, including contributions from @kdlawson (thanks for your help!)
I'll leave it to @york-stsci to merge this to stable in coordination with @ojustino 


@ojustino outstanding job throughout all these changes, congratulation in this major contribution!  and good job on getting this PR compliant with all our tests. 

For the sake of addressing the list of check boxes above, other PR will include documentation supporting these changes

Good job everyone!

york-stsci (approved):
With all tests passing, and given the comment on the CI workflow changes (and my own experience with it), this looks good to go.

PR #163, opened 2026-08-07
Linting with ruff

mperrin:
- Added ruff config to pyproject.toml
- Many minor linting and code cleanup changes using ruff.  No functional changes to STPSF are intended.
- This is _intentionally_ only a partial & incomplete linting, an incremental change. In particular the files being revised for Roman in #162 are intentionally not touched here, to avoid having to deal with conflicting edits when merging.

### DIFF
--- a/stpsf/tests/test_distortion.py
+++ b/stpsf/tests/test_distortion.py
@@ -1,9 +1,8 @@
-import pytest
 import numpy as np
+import pytest
 from astropy.io import fits
 
-from .. import distortion
-from .. import stpsf_core
+from .. import distortion, stpsf_core
 
 
 def test_apply_distortion_skew():
@@ -167,6 +166,7 @@ def test_distortion_linear_anisotropy():
     See https://github.com/spacetelescope/stpsf/issues/148
     """
     from copy import deepcopy
+
     from scipy.interpolate import interp1d
 
     def gaussian_2d(X, Y, mu, A=1.0, sigma=1.0):
@@ -281,10 +281,10 @@ def test_distortion_pixel_coords_precisely_match_siaf(plot=False):
     assert np.allclose(ynew_idl, y_aper_idl), "Y coordinates used in distort_image ought to precisely match SIAF for this setup."
 
     if plot:
-        import matplotlib, matplotlib.pyplot as plt
+        import matplotlib.pyplot as plt
 
         #--- Compute regular grid of pixels relative to the SciRef location
-        cen = (aper_npix-1)/2+1
+        (aper_npix-1)/2+1
         regular_ideal_grid_x = (x_aper_sci - aper.XSciRef) * nrc.pixelscale
         regular_ideal_grid_y = (y_aper_sci - aper.YSciRef) * nrc.pixelscale
 
--- a/stpsf/tests/test_distortion.py
+++ b/stpsf/tests/test_distortion.py
@@ -284,7 +284,6 @@ def test_distortion_pixel_coords_precisely_match_siaf(plot=False):
         import matplotlib.pyplot as plt
 
         #--- Compute regular grid of pixels relative to the SciRef location
-        (aper_npix-1)/2+1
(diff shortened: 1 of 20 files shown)


stpsf/tests/test_distortion.py
```diff
@@ -281,10 +281,10 @@ def test_distortion_pixel_coords_precisely_match_siaf(plot=False):
     assert np.allclose(ynew_idl, y_aper_idl), "Y coordinates used in distort_image ought to precisely match SIAF for this setup."
 
     if plot:
-        import matplotlib, matplotlib.pyplot as plt
+        import matplotlib.pyplot as plt
 
         #--- Compute regular grid of pixels relative to the SciRef location
-        cen = (aper_npix-1)/2+1
+        (aper_npix-1)/2+1
```
york-stsci: Given that we weren't using the "cen" variable (at least that's what I gather from ruff removing it, is there any reason that we're still keeping this line at all?

stpsf/tests/test_miri.py
```diff
@@ -239,8 +240,8 @@ def test_miri_ifu_broadening():
     fwhm_overdist = stpsf.measure_fwhm(psf, ext='OVERDIST')
     assert fwhm_overdist > fwhm_oversamp, "IFU broadening model should increase the FWHM for the distorted extensions"
 
-    fwhm_detsamp = stpsf.measure_fwhm(psf, ext='DET_SAMP')
-    fwhm_detdist = stpsf.measure_fwhm(psf, ext='DET_DIST')
+    stpsf.measure_fwhm(psf, ext='DET_SAMP')
```
york-stsci: Same here, although in this case we might be looking for a potential crash? Either way, if we're running the function, we should use the result for something or comment on why we're not.

stpsf/tests/test_nirspec.py
```diff
@@ -128,8 +129,8 @@ def test_nrs_ifu_broadening():
     fwhm_overdist = stpsf.measure_fwhm(psf, ext='OVERDIST')
     assert fwhm_overdist > fwhm_oversamp, "IFU broadening model should increase the FWHM for the distorted extensions"
 
-    fwhm_detsamp = stpsf.measure_fwhm(psf, ext='DET_SAMP')
-    fwhm_detdist = stpsf.measure_fwhm(psf, ext='DET_DIST')
+    stpsf.measure_fwhm(psf, ext='DET_SAMP')
```
york-stsci: Same comment here.

stpsf/tests/test_opds.py
```diff
@@ -656,7 +656,7 @@ def test_get_rms_per_segment():
 
 @pytest.mark.remote_data
 def test_iec_wfe_model():
-    jwst = pytest.importorskip('jwst')  # The IEC telemetry retrieval functionality requires the JWST pipeline
+    pytest.importorskip('jwst')  # The IEC telemetry retrieval functionality requires the JWST pipeline
```
york-stsci: Here it doesn't look like we're actually using "jwst" after importing it, so should we keep this line?

stpsf/tests/test_miri.py
mperrin: ah this is a bug in the following line. That assertion should have used those variables but didn't. As written it's incorrectly a duplicate of the exact same assertion 4 lines previously. Will fix.

stpsf/tests/test_opds.py
mperrin: It's used down in the call stack. `stpsf.opds.estimate_iec_induced_wfe_at_time` eventually calls `stpsf.mast_wss.retrieve_iec_heater_telemetry` which calls `jwst.lib.engdb_tools.ENGDB_Service`.

york-stsci (approved):
Looks great!

mperrin:
I see one test failed (the docs build test), for what looks like some transient timeout while setting up the test environment. I told it to retry that failed test now.

mperrin:
All tests pass after re-running the one that glitched.

PR #166, opened 2026-09-03
Implement IPC for WFI

ojustino:
The IPC reference files are per-detector, so this implementation is more similar to NIRCam's than the MIRI-like strategy we had previously discussed of adding a WFI entry to `constants.INSTRUMENT_IPC_DEFAULT_KERNEL_PARAMETERS`.

We could avoid the need for this file by querying the information from CRDS, but I didn't implement that here since it's not currently an STPSF dependency.

I added an `ipc` directory to our `prelaunch` location with the needed extra FITS file and the script that creates it. You can also re-download my temporary Box zip file.

### DIFF
--- a/stpsf/constants.py
+++ b/stpsf/constants.py
@@ -393,7 +393,6 @@ INSTRUMENT_DETECTOR_CHARGE_DIFFUSION_DEFAULT_PARAMETERS = {
     'MIRI': 0.001,  # Fit by Marshall + Marcio to ePSFs, after adding IPC
     #  0.070 Based on user reports, see issue #674. However, this was before adding IPC effects
     'WFI': 0.033,  # 20250919 Value from Goddard recommendation (stpsf issue #111)
-    'ROMANCORONAGRAPH': 0.0
 }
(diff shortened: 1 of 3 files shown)


marcio-melendez (commented):
Currently, the default value for the IPC is False
Otherwise, when IPC effects are included all the header information is updated accordingly and the effects are applied, see figure with difference between IPC and no-IPC. 

Image

ojustino:
Do you prefer to leave IPC on by default? I guess that would be more consistent with what we do for charge diffusion and jitter for the `*DIST` extensions of the PSF HDUList.

marcio-melendez:
> Do you prefer to leave IPC on by default? I guess that would be more consistent with what we do for charge diffusion and jitter for the `*DIST` extensions of the PSF HDUList.

Yes, I think it make sense to have it True as a default, it si also consistent to JWST.

marcio-melendez (approved):
I review the functional changes and with the default value as true (similar to jwst), I am hapy with this implementation. I am approving now give that all checks are passing after rebasing

PR #170, opened 2026-09-21
Roman 2 3 0 doc 1

marcio-melendez:
This is the Roman document that capture the methodology, analyses and comparisons behind the post-SCIPA data delivery from NASA. This falls under STPSF v2.3.0


marcio-melendez:
@ojustino , here is my documentation page associated with the  methodology, analyses and comparisons behind the post-SCIPA data delivery from NASA. You can add your changes to this PR or on a separate one. 

@york-stsci this PR, and possible another document-only PR, are the last PRs before the official release this week.

mperrin:
@marcio-melendez I can review this within the next ~day or two. Before that I have one question for context. I see the title of this PR is "doc 1" - does that imply there will be some additional PR(s) "doc 2" etc with additional content beyond this, too? In particular it looks like the `roman.rst` page is not modified in this PR, and I would expect that needs updates?

marcio-melendez:
Yes, roman.rst, needs to change. Justin is working on updating some of the current files to match this latest implementation. I expect for those files to have cross-references to this post-scipa page. I think Justin changes are more extensive, so that's why I label this as 1 to have the flexibility of having a larger PR with the remaining changes to the existing documents.

mperrin:
Hmmm OK. It's a little tricky to review this in the absence of that other docs PR, because as you say they will be a bit overlapping and cross referencing. 

Looking at the ipynb docs page I think there are at least 3 different kinds of info present in it. I want to suggest we consider splitting this notebook up into multiple pages.  Here are I think the distinct kinds of information:
1. detailed "for the record, here's how we did this" description of the choices made in making the Roman updates. (This is  likely to be of interest to few readers; no one else is likely to try to redo this analysis themselves, right? Good to have for the record but sort of a side stop from usage documentation.)
2. Descriptions of the WFI PSF properties, such as the field dependence. This is likely to be of broader interest and is informative to people who are using STPSF for Roman.  
3. The important note about pupil orientation, which is an important documentation statement for anyone who may be looking at those pupil files directly, which we can expect some users may do. 

I think splitting that page up will make it easier to highlight the latter two items, which I think are worth making visible to more users, and right now are potentially kind of buried towards the bottom of a very long single page.

marcio-melendez:
Agreed. It is a long document and it make sense to separate some of the sections.  I'll work on that in coordination with Justin.

marcio-melendez:
Maybe the pupil orientation section, can be added to roman.rst, instead of a separate page (@ojustino ). That means that the notebook in this PR can be separated  in two, with a "this is how we did it" and "this is what the Roman multi-field PSF looks like", giving priority to the latter (page visibility).

mperrin:
👍 

Some other thoughts for edits during that reorganization: 

- I think it's fine to have more than just 2 pages. We have at least a dozen pages for JWST...

- "SCIPA" and "post-SCIPA" are pretty jargon-y, and will be unfamiliar terms to most Roman observers. I think we need to define these, and explain some context. My general thought when writing docs is we should imagine our reader to be a grad student who is not familiar with the details of I&T on any of our missions...

- A key point I think we need to make really clear to readers is that the decomposition into a high-spatial-frequency map component and a Zernike component, that's not anything real or physical about the observatory. Rather it is a convenient parameterization. In fact the plots that we make are kind of misleading, because it looks like the Zernike component is added _after_ the "OTE exit pupil" (should be "OTA" for Roman, minor note); that's not physically correct, and really it is the case that the Zernike aberrations arise in the PM+SM etc optics in the OTA.  The way this is divided up in the software doesn't map cleanly to the actual optical physics. That's OK I guess, it doesn't have to, but I think we run the risk of confusing some users unnecessarily unless we have some clear obvious documentation of this. 

- For the JWST version of this, we don't have any comparably detailed "this is how we made the data files, and why we made those decisions". It's fine to have that added level of detail for Roman, I'm just pointing out that for JWST we instead chose to just write like "we took the phase maps from ISIM CV3 from JWST-RPT-[number]" without putting all the details in the docs. Either way OK they're just different. 

- For focus, since that's the only wavelength dependence, it would be nice to have a plot that shows how the defocus varies with wavelength. We have a [plot like that for the NIRCam defocus model](file:///Users/mperrin/Dropbox%20(Personal)/Documents/software/stpsf/docs/_build/html/jwst.html#wavelength-dependent-focus-variations). This would accompany figure 3 which shows the spatial dependence nicely but is not intuitive to show the wavelength dependence. 

- Probably somewhere we should link to the RDox page about coordinate systems, since that helps define some key framework for all this.

marcio-melendez:
In response to the comments above, among other things, I moved  the post-scipa phase map description into the Developer Reference section. The Roman PSF characterization  is a separate section under Roman simulations.

Issue #171, opened 2026-09-22
Broken link in docs


emolter:
On this page https://stpsf.readthedocs.io/en/latest/psf_grids.html#Example-PSF-grid where it says "[Photutils ePSF model](https://photutils.readthedocs.io/en/stable/epsf.html)" that link goes to a 404 not found

mperrin:
Thanks @emolter . Fix in #172

PR #172, opened 2026-09-22
fix stale link to ePSF docs in Photutils. Fixes #171

mperrin:
Very simple change: Update one URL in the docs, to accommodate a change in the photutils docs.


marcio-melendez (approved):
Looks good.

